PR #58, opened 2018-04-30
added filter for already defined endpoint options (#28)

lhein:


apupier:
I think you forgot about attributes that can be provided several times. (and I forgot to add it again when recreating the task in the new github repo) See this attribute https://github.com/camel-tooling/camel-tooling-common/blob/c04c32c1c08dc3ddc951393c456ec23cc93339b1/src/main/java/com/github/cameltooling/model/EndpointOptionModel.java#L30

apupier:
I created https://github.com/camel-tooling/camel-language-server/issues/60 to handle it

PR #85, opened 2018-06-04
fix scripts to enable clean releases

lhein:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: java
+before_install:
+- "export IS_SNAPSHOT=`grep '<version>' pom.xml -m 1 | cut -d'>' -f 2 | cut -d'<' -f 1 | cut -d'-' -f 2`"
 install: 
 - mvn install -P !build-extras -DskipTests=true -Dmaven.javadoc.skip=true -B -V
 script: mvn test -P !build-extras -B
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -12,19 +14,21 @@ env:
   global:
   - GPG_DIR="`pwd`/cd"
 after_success:
-- if [[ $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRANCH == "master" ]]; then 
+- if [[ $IS_SNAPSHOT == "SNAPSHOT" ]]; then
```
apupier: why IS_SNAPSHOT ans SNAPSHOT is used? what's the difference?

lhein: it makes it easier for me as a non-bash-nerd to work with the variables. IS_SNAPSHOT contains "SNAPSHOT" if its a snapshot version while SNAPSHOT is set to TRUE or FALSE depending whats in IS_SNAPSHOT :D

apupier: what do you think to change variable name to try to have it clearer?
For instance:
"IS_SNAPSHOT" to contain true/false
"VERSION_SUFFIX" to contain "SNAPSHOT"/*<no value>*

lhein: I could do that but that will also require the changes on common.

PR #98, opened 2018-07-02
93 support position finding of camel u ri in java file

apupier:
I recommend to review https://github.com/camel-tooling/camel-language-server/pull/97 first

please note that clients needs to be updated to have a visible effect to users (PR coming)

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperFactory.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperFactory.java
@@ -1,0 +1,42 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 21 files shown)


src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperFactory.java
```diff
@@ -0,0 +1,42 @@
...
+			if (javaParser.getCorrespondingMethodName(textDocumentItem, line) != null) {
+				return javaParser;
+			}
+		}
+		return null;
+	}
+	
+	private boolean isCamelJavaDSL(TextDocumentItem textDocumentItem, String uri) {
+		//improve this method to provide better heuristic to detect if it is a Camel file or not
+		return uri.endsWith(".java") && textDocumentItem.getText().contains("camel");
```
lhein: wouldn't it be better to check for RouteBuilder or something similar?

apupier: that is the question.
I really don't know what is the best heuristic.
If there is RouteBuilder, there will the package imported which is containing camel so i'm covering a larger scope with just "camel" than "RouteBuilder"

lhein: yeah, but you would also enable it if one class has a camel term in it....like a util class called CamelCaseUtils :) which has nothing to do with camel....so maybe better check for org.apache.camel in import section.

apupier: it is the risk. it will be enabled if there is "camel" written somewhere and that there are methods named from or to which have String as first parameter.
I think it won't give too many false-positive, especially as it means that the users have installed the Apache Camel Support.
The best way would be to have access to the full Java AST but it requires a lot more work or a bit more work and to be bound to jdt.ls

PR #104, opened 2018-07-03
Suppress warning for Sonar rule on Generic wildcard types

apupier:
it it a "false-positive" in our context because we need to have it to be
compatible with the API of a dependency.
see https://groups.google.com/d/msg/sonarqube/BLXAdLEL9QE/UuErZS6oAgAJ


apupier:
@lhein unless you have a better idea on how to write it without issues, i think we need to ignore this one and consider it as a false-positive as the Sonar guys indicated in his last mail

lhein:
no better idea for now

Issue #30, opened 2018-04-20
Provide validation of Camel URI with XML DSL


apupier:
maybe worth having a look to camel-validation-plugin and see if we can reuse it directly

apupier:
i think that as a first iteration if validation is done on save that will be enough

PR #109, opened 2018-07-11
Provide diagnostic for endpoints for XML DSL on save (#30)

apupier:
image

image

requires https://github.com/camel-tooling/camel-tooling-common/pull/26 to be reviewed and merged

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/diagnostic/DiagnosticService.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/diagnostic/DiagnosticService.java
@@ -1,0 +1,145 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.cameltooling.lsp.internal.diagnostic;
+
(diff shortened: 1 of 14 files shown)


src/main/java/com/github/cameltooling/lsp/internal/diagnostic/DiagnosticService.java
```diff
@@ -0,0 +1,145 @@
...
+		Map<CamelEndpointDetailsWrapper, EndpointValidationResult> endpointErrors = computeCamelErrors(params);
+		List<Diagnostic> diagnostics = converToLSPDiagnostics(endpointErrors, camelLanguageServer.getTextDocumentService().getOpenedDocument(params.getTextDocument().getUri()));
+		PublishDiagnosticsParams diagnosticParam = new PublishDiagnosticsParams(params.getTextDocument().getUri(), diagnostics);
+		camelLanguageServer.getClient().publishDiagnostics(diagnosticParam);
+	}
+
+	private Map<CamelEndpointDetailsWrapper, EndpointValidationResult> computeCamelErrors(DidSaveTextDocumentParams params) {
+		Map<CamelEndpointDetailsWrapper, EndpointValidationResult> endpointErrors = new HashMap<>();
+
+		if (params.getTextDocument().getUri().endsWith(".xml")) {
```
lhein: does that mean we only support validation on xml so far?

lhein:
why do we need roaster?

src/main/java/com/github/cameltooling/lsp/internal/diagnostic/DiagnosticService.java
apupier: yes.
I created this epic to gather all other closely related feature that would be nice to support soon: https://github.com/camel-tooling/camel-language-server/issues/111

apupier:
> why do we need roaster?

good point.
It seems that there is a problem in camel-route-parser dependency, it is using it but not declaring it as a dependency. I have a ClassNotFoundException on roaster classes if I don't add it to this pom

lhein:
Roaster is usually used for parsing java if I understand correctly.

apupier:
Roaster is used to parse both xml and java

only the roaster specific to java is declared (as provided :'( )
https://github.com/apache/camel/blob/a07781c240fa81d500ab13b6f4a85bb7d20daf28/tooling/camel-route-parser/pom.xml#L37-L43

apupier:
merging as approved but we can of course continue discussion on scope/impact/details of this PR

PR #123, opened 2018-07-27
Support deprecated information on Camel component and Component

apupier:
parameters (#195)

please note that Clients are nto supporting it yet, so nothing is
changing in UI yet...


apupier:
> is the strikethrough not supported by default in clients?

seems not yet supported, the feature has been added in one of the last version of the protocol

Issue #121, opened 2018-07-26
Support global endpoints in XML DSL


apupier:
currently it is only for "from" and "to", endpoint configured at camelcontext level are ignored

lhein:
fixed in #126

PR #125, opened 2018-08-03
add support for global endpoint uri completion #121

lhein:


src/main/java/com/github/cameltooling/lsp/internal/parser/ParserXMLFileHelper.java
```diff
@@ -42,13 +42,15 @@
...
+	private static final String URI_PARAM = "uri=\"";
+	
 	public String getCamelComponentUri(String line, int characterPosition) {
-		int uriAttribute = line.indexOf("uri=\"");
+		int uriAttribute = line.indexOf(URI_PARAM);
 		if(uriAttribute != -1) {
-			int nextQuote = line.indexOf('\"', uriAttribute + 5);
-			if (isBetween(characterPosition, uriAttribute + 5, nextQuote)) {
-				return line.substring(uriAttribute + 5, nextQuote);
+			int firstQuote = line.indexOf('\"', uriAttribute);
```
apupier: this is dangerous, if you have another attribute on the same line, for instance the id before, it won't work. you need to keep the search on uri=\"

if it has not break it worth adding a test for this case (or at least to raise an issue to add it)

lhein: I don't think you are right. I am starting to search in the line at the position of the URI attribute which will most likely be followed by ="value".

apupier: my bad, you're right

lhein:
created #129 to handle global endpoints in Java DSL

lhein:
@apupier this one has to be merged first

lhein:
see https://github.com/camel-tooling/camel-language-server/pull/126/files#diff-6972c603d082387f14d4b7d257cf14fcL29

apupier:
> see https://github.com/camel-tooling/camel-language-server/pull/126/files#diff-6972c603d082387f14d4b7d257cf14fcL29

needs to be reported in this PR then

lhein:
reported it

PR #126, opened 2018-08-03
Support single quotes 124

lhein:


src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelper.java
```diff
@@ -26,7 +26,8 @@
 
 public abstract class ParserFileHelper {
 	
-	protected static final List<String> CAMEL_POSSIBLE_TYPES = Arrays.asList("to", "from");
+	protected static final List<String> CAMEL_POSSIBLE_TYPES = Arrays.asList("to", "from", "endpoint");
+	protected static final List<Character> POSSIBLE_URI_CLOSURE_CHARS = Arrays.asList('\"', '\'');
```
apupier: shouldn't be better placed in the ParserXMLFileHelper? The Java xml parser is not using it right?

lhein: right

lhein:
this PR will also split the list of possible types for completion into the subclasses of ParserFileHelper.

Issue #69, opened 2018-05-28
Include Camel attribute documentation on hover


apupier:
Currently, it contains only the component documentation, it would be nice to provide the Camel attribute documentation also

PR #131, opened 2018-08-09
display hover doc for the element under the cursor #69

lhein:

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
@@ -23,6 +23,8 @@ import java.util.concurrent.CompletableFuture;
 import org.apache.camel.catalog.CamelCatalog;
 import org.eclipse.lsp4j.CompletionItem;
 
+import com.github.cameltooling.model.ComponentModel;
+
 /**
  * For a Camel URI "timer:timerName?delay=10s", it represents "delay=10s"
  *
@@ -77,5 +79,10 @@ public class OptionParamURIInstance extends CamelUriElementInstance {
 	
 	protected CamelURIInstance getCamelUriInstance() {
 		return camelURIInstance;
+	}
+	
+	@Override
+	public String getDescription(ComponentModel componentModel) {
+		return key.getDescription(componentModel);
 	}
 }
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
@@ -24,12 +24,15 @@ import org.apache.camel.catalog.CamelCatalog;
 import org.eclipse.lsp4j.CompletionItem;
 
 import com.github.cameltooling.model.ComponentModel;
+import com.github.cameltooling.model.EndpointOptionModel;
 
 /**
  * For a Camel URI "timer:timerName?delay=10s", it represents "delay=10s"
  *
  */
 public class OptionParamURIInstance extends CamelUriElementInstance {
+
+	public static final String INVALID_URI_OPTION = "The parameter you entered is not a known parameter of the component for the used Camel version!"; 
 	
 	private OptionParamKeyURIInstance key;
 	private OptionParamValueURIInstance value;
@@ -83,6 +86,7 @@ public class OptionParamURIInstance extends CamelUriElementInstance {
 	
 	@Override
 	public String getDescription(ComponentModel componentModel) {
-		return key.getDescription(componentModel);
+		EndpointOptionModel model = componentModel.getEndpointOption(getKey().getKeyName());
+		return model != null ? model.getDescription() : INVALID_URI_OPTION;
 	}
 }
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
@@ -32,7 +32,7 @@ import com.github.cameltooling.model.EndpointOptionModel;
  */
 public class OptionParamURIInstance extends CamelUriElementInstance {
 
-	public static final String INVALID_URI_OPTION = "The parameter you entered is not a known parameter of the component for the used Camel version!"; 
+	public static final String INVALID_URI_OPTION = "The parameter %s is not a known parameter of the component. Check if it is written correctly and is supported by the used Camel version (see pom.xml)."; 
 	
 	private OptionParamKeyURIInstance key;
 	private OptionParamValueURIInstance value;
@@ -87,6 +87,6 @@ public class OptionParamURIInstance extends CamelUriElementInstance {
 	@Override
 	public String getDescription(ComponentModel componentModel) {
 		EndpointOptionModel model = componentModel.getEndpointOption(getKey().getKeyName());
-		return model != null ? model.getDescription() : INVALID_URI_OPTION;
+		return model != null ? model.getDescription() : String.format(INVALID_URI_OPTION, getKey().getKeyName());
 	}
 }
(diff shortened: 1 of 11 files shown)


src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
```diff
@@ -66,12 +66,21 @@ public void testDontProvideDocumentationOnHoverForBadPlaces() throws Exception {
...
+		CamelLanguageServer camelLanguageServer = initializeLanguageServer("<from uri=\"ahc:httpUri?bufferSize=1024&\" xmlns=\"http://camel.apache.org/schema/spring\"></from>\n");
 		
 		TextDocumentPositionParams position = new TextDocumentPositionParams(new TextDocumentIdentifier(DUMMY_URI+".xml"), new Position(0, 15));
 		CompletableFuture<Hover> hover = camelLanguageServer.getTextDocumentService().hover(position);
 		
 		assertThat(hover.get()).isNull();
 	}
 
+	@Test
+	public void testProvideParameterDocumentationOnHoverWhenEndingWithAnd() throws Exception {
```
apupier: test name is not corresponding to what is tested

src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
```diff
@@ -66,12 +66,21 @@ public void testDontProvideDocumentationOnHoverForBadPlaces() throws Exception {
 	
 	@Test
 	public void testDontProvideDocumentationOnHoverWhenEndingWithAnd() throws Exception {
-		CamelLanguageServer camelLanguageServer = initializeLanguageServer("<from uri=\"ahc:httpUri?test=test&\" xmlns=\"http://camel.apache.org/schema/spring\"></from>\n");
+		CamelLanguageServer camelLanguageServer = initializeLanguageServer("<from uri=\"ahc:httpUri?bufferSize=1024&\" xmlns=\"http://camel.apache.org/schema/spring\"></from>\n");
```
apupier: why it has been changed?
it should work also when there is one attribute which is unknown

apupier:
for the PathParam, it provides hover documentation only when it is still the value predefined in the scheme. If the value has been changed, there is no more hover

for instance "timer:timerName", with cursor above timerName will provide doc in hover
"timer:myTimerName" with cursor above myTimerName will not provide doc in hover

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/PathParamURIInstance.java
```diff
@@ -86,4 +87,14 @@ private String getFilter() {
...
 	}
+	
+	@Override
+	public String getComponentName() {
+		return uriInstance.getComponentName();
+	}
+	
+	@Override
+	public String getDescription(ComponentModel componentModel) {
+		return componentModel.getEndpointOption(value).getDescription();
```
apupier: I think it is here that you need to check for non-nullity to avoid NPE when an unknown attribute is used

src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
```diff
@@ -74,4 +75,43 @@ public void testDontProvideDocumentationOnHoverWhenEndingWithAnd() throws Except
...
+	}
+	
+	@Test
+	public void testProvideSyntaxForPathParameterOnHover() throws Exception {
+		CamelLanguageServer camelLanguageServer = initializeLanguageServer("<from uri=\"kafka:fl\" xmlns=\"http://camel.apache.org/schema/spring\"></from>\n");
+		
+		TextDocumentPositionParams position = new TextDocumentPositionParams(new TextDocumentIdentifier(DUMMY_URI+".xml"), new Position(0, 19));
+		CompletableFuture<Hover> hover = camelLanguageServer.getTextDocumentService().hover(position);
+		
+		assertThat(hover.get().getContents().getLeft().get(0).getLeft()).isNotNull();
```
apupier: what about checking for the exact value instead of just non-null?

src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
```diff
@@ -74,4 +75,43 @@ public void testDontProvideDocumentationOnHoverWhenEndingWithAnd() throws Except
...
+	}
+	
+	@Test
+	public void testProvideSyntaxForEmptyPathParameterOnHover() throws Exception {
+		CamelLanguageServer camelLanguageServer = initializeLanguageServer("<from uri=\"kafka:\" xmlns=\"http://camel.apache.org/schema/spring\"></from>\n");
+		
+		TextDocumentPositionParams position = new TextDocumentPositionParams(new TextDocumentIdentifier(DUMMY_URI+".xml"), new Position(0, 17));
+		CompletableFuture<Hover> hover = camelLanguageServer.getTextDocumentService().hover(position);
+		
+		assertThat(hover.get().getContents().getLeft().get(0).getLeft()).isNotNull();
```
apupier: what about checking for the exact value instead of just non-null?

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
```diff
@@ -23,11 +23,16 @@
...
+import com.github.cameltooling.model.ComponentModel;
+import com.github.cameltooling.model.EndpointOptionModel;
+
 /**
  * For a Camel URI "timer:timerName?delay=10s", it represents "delay=10s"
  *
  */
 public class OptionParamURIInstance extends CamelUriElementInstance {
+
+	public static final String INVALID_URI_OPTION = "The parameter you entered is not a known parameter of the component for the used Camel version!";
```
apupier: What do you think to include the name of the parameter which was searched for? it would help to give more context to user

apupier: I think that it is better to remove the exclamation mark. I feel like the software is laughing at me, saying that i'm really dumb to not have an exiting parameter

apupier: which Camel version is used? How to find this information?

src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
lhein: that is risky. it just needs a small change in the catalog to break this test then.

src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
lhein: that is risky. it just needs a small change in the catalog to break this test then.

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/OptionParamURIInstance.java
lhein: good idea

lhein: ok

lhein: added a hint to the pom.xml file.

src/test/java/com/github/cameltooling/lsp/internal/hover/CamelLanguageServerHoverTest.java
apupier: yes, it is the case for all other tests too.
The other way is to be able to mock/provide our own catalog.
Advantages doesn't break on version upgrade, drawbacks if we have a bad catalog or if the structure evolves we won't detect it.

lhein:
checking for the real value now

lhein:
@apupier addressed all your concerns

PR #142, opened 2018-08-17
Provide jump to definition for ref defined as component in XML DSL

apupier:
(#139)

for instance on VS Code with Ctrl+click:

image
image


apupier:
it doesn't cover Java neither the othe rpossible notation <to ref=myRefId"/>, only <to uri="ref:myRefId"/>

Issue #55, opened 2018-04-27
On Completion, insert and replace when there is something after the cursor


apupier:
for instance, there is "timer", I put the cursor after the t and choose twitter

current result is:
"twitter:kindimer"

expected result would be:
"twitter:kind"

same applies for options.

How to handle the remaining parts if more complex such as:
"timer:value?delay=1000"
if I autocomplete after the t with twitter:kind, what would be the result?
"twitter:kind"
"twitter:kind?delay=1000"
"twitter:value"

lhein:
This is a client side behaviour from what I found out. In Eclipse it is controlled by Settings for Java but for XML there seems to be no setting at all. For VSCode I haven't checked yet.

apupier:
this can be done on server side with the textEdit parameter https://microsoft.github.io/language-server-protocol/specification
![image](https://user-images.githubusercontent.com/1105127/44076861-025a6efe-9fa2-11e8-93dc-a90e1f9c2050.png)

tsedmik:
verified with master branch

PR #143, opened 2018-08-22
refactor instance model to better reflect uri parts and allow more pr…

lhein:
…ecise position calculation

Signed-off-by: Lars Heinemann <lhein.smx@gmail.com>

PR #145, opened 2018-08-22
replace on completion (#55)

lhein:


lhein:
needs #143 to be merged first

src/main/java/com/github/cameltooling/lsp/internal/parser/ParserXMLFileHelper.java
```diff
@@ -169,7 +169,12 @@ private Document getDocumentWithLineInformation(TextDocumentItem textDocumentIte
 	@Override
 	public CamelURIInstance createCamelURIInstance(TextDocumentItem textDocumentItem, Position position, String camelComponentUri) {
 		Node correspondingCamelNode = getCorrespondingCamelNodeForCompletion(textDocumentItem, position.getLine());
-		return new CamelURIInstance(camelComponentUri, correspondingCamelNode);
+		CamelURIInstance uriInstance = new CamelURIInstance(camelComponentUri, correspondingCamelNode);
+		uriInstance.setDocument(textDocumentItem);
```
apupier: it seems that the initialization of the CamelURiInstance is duplicated than what is in the ParserJavaFileHelper.
It confirms that it should be done in the CamelURiInstance itself (as I mentioned in the preliminary PR)

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolverUtils.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+	
+	public static Map<String, Object> getCompletionResolverDataForUriInstance(CamelUriElementInstance uriInstance, String replacement) {
+		Map<String, Object> map = new HashMap<>();
+		
+		if (uriInstance != null && uriInstance.getCamelUriInstance().getAbsoluteBounds() != null) {
+			if (uriInstance.getDocument() != null) {
+				map.put(KEY_URI, uriInstance.getDocument().getUri());
+			}
+			map.put(KEY_LINE, "" + uriInstance.getCamelUriInstance().getAbsoluteBounds().getStart().getLine());
```
apupier: what do you think to use Integer.toString() instead of concatenating with an empty String?

it avoids implicit boxing but it might be less readable

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/CamelComponentAndPathUriInstance.java
```diff
@@ -0,0 +1,140 @@
...
+	 * 
+	 * @param positionInUri	the position
+	 * @return	the filter string or null if not to be filtered
+	 */
+	private String getFilter(int positionInUri) {
+		String componentName = getComponent().getComponentName();
+		if (componentName != null && componentName.trim().length()>0 && getStartPositionInUri() != positionInUri) {
+			String filter = componentName.substring(getStartPositionInUri(), positionInUri < componentName.length() ? positionInUri : componentName.length());
+			// if cursor is behind the ":" then we add it to the filter to exclude other components with the same starting chars like ahc and ahc-https
+			if (positionInUri-getStartPositionInUri() > componentName.length()) filter += ":";
```
apupier: I would prefer the 2 statements on separated lines with curly braces

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolveHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextEdit;
+
+/**
+* @author lheinema
+*/
+public class CompletionResolveHandler {
+
+	public CompletionItem resolve(CompletionItem param) {
+		Map data = CompletionResolverUtils.getDataFromJSON(param.getData(), Map.class);
```
apupier: using generic Map\<?,?> data will avoid a warning "map is a raw type"

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolveHandler.java
```diff
@@ -0,0 +1,59 @@
...
+		// clean resolve data
+		param.setData(null);
```
apupier: what is the purpose of cleaning the data on the param?

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolveHandler.java
```diff
@@ -0,0 +1,59 @@
...
+
+		if (data != null && !data.isEmpty()) {
+			int line = Integer.parseInt(data.containsKey(CompletionResolverUtils.KEY_LINE) ? (String)data.get(CompletionResolverUtils.KEY_LINE) : "0");
```
apupier: note: I was thinking to use the getOrDefault Java 8 syntax here but unfortunately the Map retrieved from GSon is not providing the generic types :'(

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolverUtils.java
```diff
@@ -0,0 +1,73 @@
...
+	private CompletionResolverUtils() {
+		// util class
+	}
+	
+	public static Map<String, Object> getCompletionResolverDataForUriInstance(CamelUriElementInstance uriInstance, String replacement) {
+		Map<String, Object> map = new HashMap<>();
+		
+		if (uriInstance != null && uriInstance.getCamelUriInstance().getAbsoluteBounds() != null) {
+			if (uriInstance.getDocument() != null) {
+				map.put(KEY_URI, uriInstance.getDocument().getUri());
```
apupier: unless I miss something, this key "KEY_URI" is not used anywhere. Maybe we can get rid of it to simplify code?

src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
```diff
@@ -41,6 +44,7 @@ public CamelComponentSchemesCompletionsFuture(String filterText) {
 					CompletionItem completionItem = new CompletionItem(componentModel.getSyntax());
 					completionItem.setDocumentation(componentModel.getDescription());
 					completionItem.setDeprecated(Boolean.valueOf(componentModel.getDeprecated()));
+					completionItem.setData(CompletionResolverUtils.getCompletionResolverDataForUriInstance(uriElement, completionItem.getLabel()));
```
apupier: why did you choose to pass Data to the CompletionItem and provide a textEdit only on resolveCompletionItem?
it seems to me that in com.github.cameltooling.lsp.internal.completion.CompletionResolveHandler.resolve(CompletionItem) it is just retrieving values from the data to create a textEdit without doing any computations. I don't see the interest then.

lhein: I was basically just imitating what JDT.LS was doing. If you think there is a better way then we can improve it later on.

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolveHandler.java
lhein: tbh I have no clue if its required. I copied that from another LSP.

src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
apupier: the difference is that in JDt they are doing a lot more things that is taking time to compute in the resolve method https://github.com/eclipse/eclipse.jdt.ls/blob/c739690a58d27a9077a0b7d2daece258ba64cb11/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/CompletionResolveHandler.java#L70
i think it is is useless for us for now.
Not doing it is greatly simplifying code, I really think it is worthy to do it in this PR

apupier:
@lhein requires rebase to solve conflicts

lhein:
applied simplification of the completion logic

lhein:
there is currently no test checking for the text edit afair

Issue #150, opened 2018-09-10
[Regression]Default values are no more added on attribute completion


apupier:
when calling completion for instance on "timer:timerName?" and choose "period"
it should end up with "timer:timerName?period=1000" but it currently completes with ""timer:timerName?period" only

apupier:
introduced with https://github.com/camel-tooling/camel-language-server/issues/55

apupier:
another related issue detected:
when the uri is empty, there is no completion

tsedmik:
verified in Eclipse (standalone) - Tooling for Apache Camel based on Apache Camel implementation of Language Server Protocol 1.0.0.201809120921

PR #151, opened 2018-09-10
Fix regression for default values completion (#150)

apupier:
- fix default values on completion regression
- fix completion with empty uri


apupier:
@lhein please review asap as it is fixing regression affecting one of the most used workflow: completion with an empty uri

Issue #158, opened 2018-09-12
Provide completion with potential ids of direct/direct-vm/vm/seda with XML DSL


apupier:
for instance, if I have a file with:
```
<from uri="direct:anId"/>

<...>

<to uri="direct:"/>
```

completion on the "to after the ':' character should provide "anId".

lhein:
merged PR

tsedmik:
Hi @lhein, there is no need to assign issues me. It is sufficient to move it to _on QA_ in planning board ;-)

tsedmik:
verified with master branch (in Eclipse IDE)

PR #160, opened 2018-09-25
added completions for existing direct, direct-vm, vm and seda endpoints #158

lhein:

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
@@ -58,7 +58,7 @@ public final class CamelComponentSchemesCompletionsFuture implements Function<Ca
 			.filter(FilterPredicateUtils.matchesCompletionFilter(filterString))
 			.collect(Collectors.toList());
 	
-		if (CompletionResolverUtils.isDirectComponentKind(uriElement)) {
+		if (CompletionResolverUtils.isReferenceComponentKind(uriElement)) {
 			addExistingEndpointsOfSameSchemeCompletionItems(result);
 		}
 		
--- a/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
@@ -22,23 +22,32 @@ import java.util.stream.Collectors;
 
 import org.apache.camel.catalog.CamelCatalog;
 import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.TextDocumentItem;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.github.cameltooling.lsp.internal.instancemodel.CamelUriElementInstance;
+import com.github.cameltooling.lsp.internal.parser.ParserXMLFileHelper;
+import com.github.cameltooling.lsp.internal.references.ReferencesProcessor;
 import com.github.cameltooling.model.util.ModelHelper;
 
 public final class CamelComponentSchemesCompletionsFuture implements Function<CamelCatalog, List<CompletionItem>> {
 	
+	private static final Logger LOGGER = LoggerFactory.getLogger(CamelComponentSchemesCompletionsFuture.class);
+	
 	private CamelUriElementInstance uriElement;
 	private String filterString;
+	private TextDocumentItem docItem;
 	
-	public CamelComponentSchemesCompletionsFuture(CamelUriElementInstance uriElement, String filterText) {
+	public CamelComponentSchemesCompletionsFuture(CamelUriElementInstance uriElement, String filterText, TextDocumentItem docItem) {
 		this.uriElement = uriElement;
 		this.filterString = filterText;
+		this.docItem = docItem;
 	}
 	
 	@Override
 	public List<CompletionItem> apply(CamelCatalog catalog) {
-		return catalog.findComponentNames().stream()
+		List<CompletionItem> result = catalog.findComponentNames().stream()
 				.map(componentName -> ModelHelper.generateComponentModel(catalog.componentJSonSchema(componentName), true))
 				.map(componentModel -> {
 					CompletionItem completionItem = new CompletionItem(componentModel.getSyntax());
@@ -49,5 +58,21 @@ public final class CamelComponentSchemesCompletionsFuture implements Function<Ca
 				})
 				.filter(FilterPredicateUtils.matchesCompletionFilter(filterString))
 				.collect(Collectors.toList());
+		
+		if (ReferencesProcessor.isDirectComponentKind(uriElement)) {
+			// find all endpoints of same component scheme and add them to the list of completion suggestions
+			try {
+				List<String> allEndpointIDsWithScheme = ReferencesProcessor.retrieveEndpointIDsOfScheme(uriElement.getComponentName(), new ParserXMLFileHelper(), docItem);
+				for (String s : allEndpointIDsWithScheme) {
+					CompletionItem completionItem = new CompletionItem(s);
+					CompletionResolverUtils.applyTextEditToCompletionItem(uriElement, completionItem);
+					result.add(completionItem);
+				}
+			} catch (Exception ex) {
+				LOGGER.error("Error retrieving existing " + uriElement.getComponentName() + " endpoints!", ex);
+			}
+		}
+		
+		return result;
 	}
 }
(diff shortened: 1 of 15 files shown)


src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
```diff
@@ -49,5 +58,21 @@ public CamelComponentSchemesCompletionsFuture(CamelUriElementInstance uriElement
 				})
 				.filter(FilterPredicateUtils.matchesCompletionFilter(filterString))
 				.collect(Collectors.toList());
+		
+		if (ReferencesProcessor.isDirectComponentKind(uriElement)) {
+			// find all endpoints of same component scheme and add them to the list of completion suggestions
```
apupier: instead of writing a comment, I think extracting a method would be better for readability

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
```diff
@@ -99,16 +100,34 @@ private boolean isReference(CamelURIInstance camelURIInstanceToSearchReference,
...
+	private static String getDirectId(CamelURIInstance camelDirectURIInstance) {
 		Set<PathParamURIInstance> pathParams = camelDirectURIInstance.getComponentAndPathUriElementInstance().getPathParams();
 		if (!pathParams.isEmpty()) {
 			return pathParams.iterator().next().getValue();
 		}
 		return null;
 	}
 
-	private boolean isDirectComponentKind(CamelURIInstance camelURIInstanceToSearchReference) {
+	public static boolean isDirectComponentKind(CamelUriElementInstance camelURIInstanceToSearchReference) {
```
apupier: if used in several places ReferencesProcessor.isDirectComponentKind doesn't make a lot of meaning to stay in ReferencesProcessor.
it would be better to extract it to a specific class

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
```diff
@@ -99,16 +100,34 @@ private boolean isReference(CamelURIInstance camelURIInstanceToSearchReference,
...
+	public static List<String> retrieveEndpointIDsOfScheme(String scheme, ParserXMLFileHelper xmlFileHelper, TextDocumentItem docItem) throws Exception {
+		List<Node> allEndpoints = xmlFileHelper.getAllEndpoints(docItem);
+		List<String> endpointIDs = new ArrayList<>();
+		for (Node endpoint : allEndpoints) {
+			String uriToParse = CamelXmlHelper.getSafeAttribute(endpoint, "uri");
+			if (uriToParse != null) {
+				CamelURIInstance uriInstance = new CamelURIInstance(uriToParse, endpoint, docItem);
+				if (isDirectComponentKind(uriInstance) && uriInstance.getComponentName().equalsIgnoreCase(scheme)) {
+					String dId = getDirectId(uriInstance);
+					String directValue = String.format("%s:%s", scheme, dId);
```
apupier: I don't understand why there is a String.format("%s:%s").
Seems like connection between direct and direct-vm won't be handled

apupier: computation of directValue can be done inside the following if statement

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
```diff
@@ -99,16 +100,34 @@ private boolean isReference(CamelURIInstance camelURIInstanceToSearchReference,
...
 		}
 		return null;
 	}
 
-	private boolean isDirectComponentKind(CamelURIInstance camelURIInstanceToSearchReference) {
+	public static boolean isDirectComponentKind(CamelUriElementInstance camelURIInstanceToSearchReference) {
 		return POSSIBLE_DIRECT_REFERENCE.contains(camelURIInstanceToSearchReference.getComponentName());
 	}
 
+	public static List<String> retrieveEndpointIDsOfScheme(String scheme, ParserXMLFileHelper xmlFileHelper, TextDocumentItem docItem) throws Exception {
```
apupier: this method is used only in CamelComponentSchemesCompletionsFuture, seems it would be a better fit there

apupier: what do you think to reuse retrieveAllEndpoints() and then search for the ids?
it will avoid to duplicate a part of the code

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
lhein: how? I use the value both in the IF and inside the IF clause

apupier: effectively, i missed the usage inside the if

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
lhein: one is static and the other isn't. It would require to make the other method static too and to introduce additional parameters for TextDocumentItem and XMLParser. Besides that I plan to relocate that method into a Util class

lhein: relocated into a util class

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
lhein: I am still wondering if you can wire direct and direct-vm. Documentation is a bit sparse on that topic.

src/main/java/com/github/cameltooling/lsp/internal/references/ReferencesProcessor.java
lhein: relocated to util class

src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentSchemesCompletionsFuture.java
lhein: sounds good

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolverUtils.java
```diff
@@ -16,18 +16,31 @@
...
 import com.github.cameltooling.lsp.internal.instancemodel.CamelUriElementInstance;
+import com.github.cameltooling.lsp.internal.instancemodel.PathParamURIInstance;
+import com.github.cameltooling.lsp.internal.parser.ParserXMLFileHelper;
 
 /**
  * @author lheinema
  */
 public class CompletionResolverUtils {
 	
+	private static final List<String> POSSIBLE_DIRECT_REFERENCE = Arrays.asList("direct","direct-vm", "seda", "vm", "amqp", "jms", "activemq", "broker");
```
apupier: I think this attribute should be renamed now that there not only "direct" and "direct-vm"
maybe POSSIBLE_REFERENCES or POSSIBLE_CROSS_REFERENCE

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolverUtils.java
```diff
@@ -48,4 +61,36 @@ public static void applyTextEditToCompletionItem(CamelUriElementInstance uriInst
 			}
 		}
 	}
+	
+	public static boolean isDirectComponentKind(CamelUriElementInstance camelURIInstanceToSearchReference) {
```
apupier: would be better to rename the method too

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolverUtils.java
```diff
@@ -48,4 +61,36 @@ public static void applyTextEditToCompletionItem(CamelUriElementInstance uriInst
...
+		return POSSIBLE_DIRECT_REFERENCE.contains(camelURIInstanceToSearchReference.getComponentName());
+	}
+
+	public static String getDirectId(CamelURIInstance camelDirectURIInstance) {
```
apupier: needs to rename this method too

src/main/java/com/github/cameltooling/lsp/internal/completion/CompletionResolverUtils.java
apupier: I think it should be available in another class. it is not used only for completion

apupier: broker is not part of camel-core, it is a component provided separately
seems hosted here http://activemq.apache.org/broker-camel-component.html

no completion for it:
image

is it really worthy to add it as we can't even provide a non-regression test for now?

apupier: activemq seems to not be part of camel-core either 

http://camel.apache.org/activemq.html

image

apupier: for JMS I'm not sure that itis the correct behavior. The "name" to reference might not be the second element.

for instance 
image
I think that the completion for jms:another is wrong because "another" is a topic although when not specified it is a queue

apupier: for amqp, it is the same than for JMS

lhein: atm it is only used for completion. lets move it once thats no longer the case.

tsedmik:
related to #158

PR #164, opened 2018-10-17
added validation for component dependencies #91

lhein:

needs the following PR merged first to work:
https://github.com/camel-tooling/camel-tooling-common/pull/30

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/util/MavenUtils.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/util/MavenUtils.java
@@ -1,0 +1,176 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.github.cameltooling.lsp.internal.util;
+
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.net.URI;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.camel.catalog.CamelCatalog;
+import org.apache.camel.catalog.JSonSchemaHelper;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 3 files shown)


lhein:
> missing tests

Yes, currently missing. It will require a fully fledged project including a pom.xml to have it tested correctly. Not sure if we have something similar in the existing tests yet.

> I think it must be specified that in commit comment (and after in changelog) that the added validation is working only if both camel file and pom file are opened (unless i misunderstood while reading the code?)

Wrong. The pom doesn't need to be opened. The pom path is determined by taking the path of the camel file as base and search in the parent folders.

> how is it working if the pom.xml relies on other maven projects available only locally? it expects to have the Client having built them and pushed in the local .m2, isn't it? 

Nothing needs to be built. I simply parse the pom.xml in the project path.

> In this case, it means that it will work only when the client and the server are on the same machine

I think that is true anyways. If the server has no physical access to both the camel file and the pom then the whole LSP will most likely not work if I understand it fully. And that is because the LSP events only send a file:/// uri to the server.

> in case, project are not built or not on the same machine I think it will cause false-positive, isn't it?

No clue what you mean here. Nothing needs to be built imo.

src/main/java/com/github/cameltooling/lsp/internal/util/MavenUtils.java
```diff
@@ -0,0 +1,176 @@
...
+		while (f.getParentFile() != null && pom == null) {
+			f = f.getParentFile();
+			pom = getPomFromPath(f);
+		}
+		return pom;
+	}
+
+	public static File getPomFromPath(File path) {
+		File[] files = path.listFiles( (File pathname) -> {
+			return pathname.getName().equalsIgnoreCase("pom.xml");
```
apupier: better to put "pom.xml" constant on the left of the equalsIgnoreCase to avoid potential NPE

apupier:
> > In this case, it means that it will work only when the client and the server are on the same machine
> 
> I think that is true anyways. If the server has no physical access to both the camel file and the pom then the whole LSP will most likely not work if I understand it fully. And that is because the LSP events only send a file:/// uri to the server.

No, the content of the opened file are opened. there is no direct access so far on the server side to the filesystem of the client side (which is most of the time the same but if we want it better distributed in the future it won't be the case)

lhein:
I am still struggling a bit with this feature. Possible issues I see:

- its not clear if we always have a maven project, could be also gradle or just plain camel file
- I haven't found a good way yet to resolve the model of the pom, just pull parsing the pom.xml is not enough
- this validation will also rely on the set camel version in the project which might be there or might be not depending on what kind of project we have
- possible performance issues when loading the camel catalog for a specific camel version

Overall I am rather unsure if the feature should be implemented in that way.

@apupier any thoughts?

apupier:
I think we should not handle Maven in our language server and delegate it to jdt.ls. Spring Tool Suite 4 is doing it this way. The drawback is that jdt.ls is required to be available to be it is easier to implement and maintain and less error-prone.
The best is to decouple in our code the calls to jdt.ls so that the feature is just deactivated (with a possible suggestion to install it) in case the jdt.ls is not available.

apupier:
closing as the PR has not been touched for almost a year

PR #187, opened 2018-11-12
153 java11 ci test

apupier:


apupier:
on PR branch:
image


on non PR branch it will do it this way:
image

Issue #147, opened 2018-09-07
Upgrade to Camel 2.23.0


apupier:
(when released)

- check positions are now more accurate for validation (tests might need to be adjusted) (see https://issues.apache.org/jira/browse/CAMEL-12640 )

tsedmik:
verified with master branch

Issue #113, opened 2018-07-11
Improve range position computation to be more precise than "just" the line in Diagnostic


apupier:

apupier:
it has been improved for Java dsl in Camel 2.23

apupier:
see https://github.com/camel-tooling/camel-language-server/issues/147

djelinek:
verified with 1.1.0-SNAPSHOT

PR #194, opened 2018-11-29
Upgrade to Camel 2.23.0 #147

apupier:
better to review and merge https://github.com/camel-tooling/camel-tooling-common/pull/37 before


apupier:
same behavior than before for validation position. There is an extra parameter available to provide a better position but it is not a range so not trivial. Better to handle in a separate enhancement https://github.com/camel-tooling/camel-language-server/issues/113

Issue #195, opened 2018-11-30
Upgrade assertj to have same version in all projects to be compatible with Infinitest


apupier:
it seems that several assertj versions are not working with assertj.

Last working version is 3.10.0

apupier:
NoSuchMethodError (org.assertj.core.api.Assertions.assertThat(Ljava/lang/String;)Lorg/assertj/core/api/AbstractStringAssert;) in 0,17.testProvideCompletionForCamelBlueprintNamespace[ahc:ht - Position 

![image](https://user-images.githubusercontent.com/1105127/49277080-b71e2f00-f480-11e8-9d99-ec033af80993.png)

djelinek:
verified with master branch

PR #196, opened 2018-11-30
Switch back to an AssertJ version compatible with InfiniTest #195

apupier:


apupier:
it is due to a mix of AssertJ versions on same workspace https://github.com/infinitest/infinitest/issues/269#issuecomment-443639326 i will upgrade in tooling common instead

apupier:
upgrade with this PR https://github.com/camel-tooling/camel-tooling-common/pull/38

PR #248, opened 2019-07-04
163 basicblind completion for kamel groovy

apupier:
# Pull Request informations

3. Green job for PR
4. Approved PR

## PR labels default process

## Tests

## PR workflow progress

2. [ ] Green **job for PR**


src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperFactory.java
```diff
@@ -20,6 +20,10 @@
 
 public class ParserFileHelperFactory {
 	
+	private static final String CAMELK_GROOVY_FILENAME_SUFFIX = ".camelk.groovy";
```
lhein: hasn't that suffix suggestion been declined in favor of the modeline?

src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperFactory.java
```diff
@@ -30,10 +34,30 @@ public ParserFileHelper getCorrespondingParserFileHelper(TextDocumentItem textDo
...
+			if (camelKGroovyDSLParser.getCorrespondingMethodName(textDocumentItem, line) != null) {
+				return camelKGroovyDSLParser;
+			}
 		}
 		return null;
 	}
 	
+	private boolean isCamelKGroovyDSL(TextDocumentItem textDocumentItem, String uri) {
+		//improve this method to provide better heuristic to detect if it is a Camel file or not
+		return uri.endsWith(CAMELK_GROOVY_FILENAME_SUFFIX)
```
lhein: same comment as above

src/test/resources/workspace/sample.camelk.groovy
```diff
@@ -0,0 +1,4 @@
+from('direct:greeting-api')
```
lhein: same comments as above

src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperFactory.java
apupier: it has not been declined completely.
They asked for the modeline but we don't reach an agreement on what is the best.
The proposal is to support 3 ways:
- modeline-like
- filename *.camelk.groovy
- shebang (this one will require more work on runtime side)

bfitzpat:
My vote is for the file extension hack vs. something inside a file, as that doesn't change the file contents. Adding a one-line marker that essentially says "hey, this is for camel-k" is not very useful to the user who already knows the file is meant for camel-k. That said, a comment in the file is more explicit than the file extension. 

I would go with the file extension if the Camel-K team is ok with it but we need to be on the same page.

apupier:
discussed in call with Brian:
- currently there is no priority on this task (completion on Camel-K files), so we would like to avoid to spend too much time for a first iteration (we already spent more times asking what to do and then discussing it than it took to implement it)
- there is no plan to have mandatory modeline or shebang (or file extension) in future.
- the current PR is letting users choose the way they want, they can use both (extension or modeline-like). Everything stays open for the future.
- the modification is almost non-intrusive, there are very low chance that someone using a *.camelk.groovy file with a from/to method inside and having Camel extension installed is writing a file which is not a Camel-K file

given all of that we are thinking that we should merge to allow people play with it and if it comes with higher priority, we can rediscuss and rework it

Issue #264, opened 2019-08-08
Provide completion for keywords after "camel." key in application.properties


apupier:
it will allow to have a first smaller step going to full components completion https://github.com/camel-tooling/camel-language-server/issues/239

- should provide only for application.properties (avoid collision with Camel Kafka connect properties file)
- be able to work on keys of properties File

PR #265, opened 2019-08-08
provide initial top-level completion for application.properties #264

apupier:
image

image

- it allows to setup the mechanism for application.properties
- it has hardcoded 3 values only for now
- includes a small refactor to extract some useful methods of
ParserHelper

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/CamelTextDocumentService.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/CamelTextDocumentService.java
@@ -61,6 +61,7 @@ import org.eclipse.lsp4j.services.TextDocumentService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.cameltooling.lsp.internal.completion.CamelApplicationPropertiesCompletionProcessor;
 import com.github.cameltooling.lsp.internal.completion.CamelEndpointCompletionProcessor;
 import com.github.cameltooling.lsp.internal.definition.DefinitionProcessor;
 import com.github.cameltooling.lsp.internal.diagnostic.DiagnosticService;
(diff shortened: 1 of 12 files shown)


apupier:
@davsclaus I put only 3 top-level for now main/component/rest
Do we have a full list of what is possible just after `camel.`? Unless i missed it, i don't think it is part of CamelCatalog (or it is from Camel 3 catalog? in this case will be more work as we don't support several version yet)
Do we have some description somewhere for these "top level" elements to guide users?

bfitzpat:
Is there some documentation about this application.properties file somewhere we can look at?

apupier:
> Is there some documentation about this application.properties file somewhere we can look at?

@davsclaus the question is for you :-)

davsclaus:
@apupier you have the top level groups defined in this json file
https://github.com/apache/camel/blob/master/core/camel-main/src/main/resources/META-INF/camel-main-configuration-metadata.json

davsclaus:
And yeah you have all the options, and documentation in that json file (its identical to a spring boot metadata file for its tooling) - but we generate it oursleves so the format can be updated/changed etc.

davsclaus:
And then you have all the Camel components that start with 

camel.component.xxx

where xxx is the component name, and its options are what's in the camel-catalog today.

apupier:
> @apupier you have the top level groups defined in this json file
> https://github.com/apache/camel/blob/master/core/camel-main/src/main/resources/META-INF/camel-main-configuration-metadata.json

i'm a bit surprised that this file contains `main`/`hystrix`/`rest ` but not `component`,  why the first 3 but not the last one?

> And yeah you have all the options, and documentation in that json file (its identical to a spring boot metadata file for its tooling) - but we generate it oursleves so the format can be updated/changed etc.

Are these options really "SprintBoot options"? or it is really only a trick to leverage SpringBoot tooling and these options are ignored by SpringBoot at runtime?

apupier:
> And then you have all the Camel components that start with
> 
> camel.component.xxx
> 
> where xxx is the component name

I plan to handle it in a different PR https://github.com/camel-tooling/camel-language-server/pull/267

davsclaus:
The camel-main only has the options that it provides, eg options you can configure on camel itself (camel main), each component has their own set of files. And then you have camel-catalog which has all of them for a given camel release.

And yeah its a little trick so tooling like vscode can do auto completion because it sees this json file and think its a spring boot project.

apupier:
> you have the top level groups defined in this json file
> https://github.com/apache/camel/blob/master/core/camel-main/src/main/resources/META-INF/camel-main-configuration-metadata.json

ok so I'm missing the hystrix group.
@davsclaus Why the "component" is not listed in this file?

apupier:
updated PR with all top-level groups:
Screenshot from 2019-09-03 14-22-08

davsclaus:
All the components are in the spring-boot starter JARs or in the camel-catalog.

Issue #284, opened 2019-09-03
Provide support for Yaml files


apupier:
2 use cases (might be splitted in 2 different PRs):

- Camel-K standalone file https://github.com/apache/camel-k-runtime/blob/master/examples/camel-k-runtime-example-yaml/src/main/resources/routes.yaml
- under the hood CRD https://docs.google.com/presentation/d/1jXMuHtRvcgzUORiYTkQUslCyhBU_YuYR9LJSLWDsooo/edit#slide=id.g585e82468f_0_0
- KNative CRD (pretty sure I have seen a specific Knative CRD somewhere but cannot find references)

heuristic to provide the completion to be determined, some proposals:
- for standalone file : filename *.camelk.yaml/modeline like and a from/to inside steps and followed by an uri? (format is not yet fully determined as far as I know maybe we can support the different ones) https://docs.google.com/presentation/d/1jXMuHtRvcgzUORiYTkQUslCyhBU_YuYR9LJSLWDsooo/edit#slide=id.g5fe2be6cc0_2_64
- maybe api version which is on the first line and then the from/to from inside a sources/sinks?

Will require to add the filetype association in each client IDE

lhein:
feature has been implemented @djelinek @tsedmik

apupier:
found reference to the Knative CRD https://github.com/knative/eventing-contrib/blob/master/camel/source/samples/
it seems to work with it too. Maybe will worth to have test for this case too

apupier:
seems that the initial iteration is working fine.
closing this issue. We can open more specific issue.

PR #291, opened 2019-09-10
adding yaml file support #284

lhein:

# Pull Request informations

3. Green job for PR
4. Approved PR

## PR labels default process

## Tests

- [x] Are there **Unit tests**?

## PR workflow progress

1. [x] Tagged with relevant **PR labels**
2. [x] Green **job for PR**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -305,6 +305,11 @@
 			<version>${awaitility.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.21</version>            
+		</dependency>
 	</dependencies>
 	
 	<repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -308,7 +308,7 @@
 		<dependency>
 			<groupId>org.yaml</groupId>
 			<artifactId>snakeyaml</artifactId>
-			<version>1.21</version>            
+			<version>1.25</version>            
 		</dependency>
 	</dependencies>
 	
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/YamlDSLModelHelper.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/YamlDSLModelHelper.java
@@ -1,0 +1,37 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.cameltooling.lsp.internal.instancemodel;
+
+public class YamlDSLModelHelper implements DSLModelHelper {
+	
+	private String line;
+
+	public YamlDSLModelHelper(String line) {
+		this.line = line;
+	}
+
+	@Override
+	public String getParametersSeparator() {
+		return "&";
+	}
+
+	@Override
+	public String getTypeDeterminingProducerConsumer() {
+		return line;
+	}
+	
+}
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/YamlDSLModelHelper.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/YamlDSLModelHelper.java
@@ -18,10 +18,10 @@ package com.github.cameltooling.lsp.internal.instancemodel;
 
 public class YamlDSLModelHelper implements DSLModelHelper {
 	
-	private String line;
+	private String type;
 
-	public YamlDSLModelHelper(String line) {
-		this.line = line;
+	public YamlDSLModelHelper(String type) {
+		this.type = type;
 	}
 
 	@Override
@@ -31,7 +31,7 @@ public class YamlDSLModelHelper implements DSLModelHelper {
 
 	@Override
 	public String getTypeDeterminingProducerConsumer() {
-		return line;
+		return type;
 	}
 	
 }
--- a/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
@@ -30,7 +30,7 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 	public static final String STEP_KEY = "steps:";
 	public static final String URI_KEY = "uri:";
 	public static final String REST_KEY = "rest:";
-	public static final String FROM_KEY = "- from:";
+	public static final String FROM_KEY = "from:";
 	public static final String TO_KEY = "- to:";
 
 	@Override
@@ -38,11 +38,11 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 		String camelComponentURI = null;
 		String trLine = line.trim();
 		if (trLine.startsWith(URI_KEY) && URI_KEY.length() < characterPosition) {
-			camelComponentURI = extractCamelCOmponentURI(trLine, URI_KEY);
+			camelComponentURI = extractCamelComponentURI(trLine, URI_KEY);
 		} else if(trLine.startsWith(FROM_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
-			camelComponentURI = extractCamelCOmponentURI(trLine, FROM_KEY);
+			camelComponentURI = extractCamelComponentURI(trLine, FROM_KEY);
 		} else if(trLine.startsWith(TO_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
-			camelComponentURI = extractCamelCOmponentURI(trLine, TO_KEY);
+			camelComponentURI = extractCamelComponentURI(trLine, TO_KEY);
 		}
 		return camelComponentURI;
 	}
@@ -50,14 +50,14 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 	@Override
 	public CamelURIInstance createCamelURIInstance(TextDocumentItem textDocumentItem, Position position,
 			String camelComponentUri) {
-		CamelURIInstance uriInstance = new CamelURIInstance(camelComponentUri, new YamlDSLModelHelper(getCorrespondingLine(textDocumentItem, position.getLine())), textDocumentItem);
+		CamelURIInstance uriInstance = new CamelURIInstance(camelComponentUri, new YamlDSLModelHelper(getCorrespondingType(textDocumentItem, position.getLine())), textDocumentItem);
 		int start = getStartCharacterInDocumentOnLinePosition(textDocumentItem, position);
 		uriInstance.setStartPositionInDocument(new Position(position.getLine(), start));
 		uriInstance.setEndPositionInDocument(new Position(position.getLine(), start+camelComponentUri.length()));
 		return uriInstance;
 	}
 
-	private String extractCamelCOmponentURI(String trLine, String key) {
+	private String extractCamelComponentURI(String trLine, String key) {
 		return trLine.substring(trLine.indexOf(key)+ key.length()).trim().replaceAll("\"", "");
 	}
 
@@ -72,7 +72,7 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 		return position.getCharacter() - line.indexOf('"')-1;
 	}
 
-	public String getCorrespondingLine(TextDocumentItem textDocumentItem, int lineNumber) {
+	public String getCorrespondingType(TextDocumentItem textDocumentItem, int lineNumber) {
 		for (int lineNo = lineNumber; lineNo >=0; lineNo--) {
 			String tempLine = parserFileHelperUtil.getLine(textDocumentItem, lineNo).trim();
 			if (tempLine.startsWith(TO_KEY)) {
@@ -80,7 +80,7 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 			} else if (tempLine.startsWith(FROM_KEY)) {
 				return "from";
 			} else if (tempLine.startsWith(REST_KEY)) {
-				return null;			
+				return null;
 			}
 		}
 		return null;
--- a/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
@@ -1,0 +1,84 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.cameltooling.lsp.internal.parser;
+
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.TextDocumentItem;
+
+import com.github.cameltooling.lsp.internal.instancemodel.CamelURIInstance;
+import com.github.cameltooling.lsp.internal.instancemodel.YamlDSLModelHelper;
+
+/**
+ * @author Lars Heinemann
+ */
+public class CamelKYamlDSLParser extends ParserFileHelper {
+	
+	public static final String STEP_KEY = "steps:";
+	public static final String URI_KEY = "uri:";
+	public static final String REST_KEY = "rest:";
+	public static final String FROM_KEY = "- from:";
+	public static final String TO_KEY = "- to:";
+
+	@Override
+	public String getCamelComponentUri(String line, int characterPosition) {
+		String camelComponentURI = null;
+		String trLine = line.trim();
+		if (trLine.startsWith(URI_KEY) && URI_KEY.length() < characterPosition) {
+			camelComponentURI = trLine.substring(trLine.indexOf(URI_KEY)+ URI_KEY.length()).trim().replaceAll("\"", "");
+		} else if(trLine.startsWith(FROM_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
+			camelComponentURI = trLine.substring(trLine.indexOf(FROM_KEY)+ FROM_KEY.length()).trim().replaceAll("\"", "");
+		} else if(trLine.startsWith(TO_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
+			camelComponentURI = trLine.substring(trLine.indexOf(TO_KEY)+ TO_KEY.length()).trim().replaceAll("\"", "");
+		}
+		return camelComponentURI;
+	}
+
+	@Override
+	public CamelURIInstance createCamelURIInstance(TextDocumentItem textDocumentItem, Position position,
+			String camelComponentUri) {
+		CamelURIInstance uriInstance = new CamelURIInstance(camelComponentUri, new YamlDSLModelHelper(getCorrespondingLine(textDocumentItem, position.getLine())), textDocumentItem);
+		int start = getStartCharacterInDocumentOnLinePosition(textDocumentItem, position);
+		uriInstance.setStartPositionInDocument(new Position(position.getLine(), start));
+		uriInstance.setEndPositionInDocument(new Position(position.getLine(), start+camelComponentUri.length()));
+		return uriInstance;
+	}
+
+	private int getStartCharacterInDocumentOnLinePosition(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return line.indexOf('"') + 1;
+	}
+
+	@Override
+	public int getPositionInCamelURI(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return position.getCharacter() - line.indexOf('"')-1;
+	}
+
+	public String getCorrespondingLine(TextDocumentItem textDocumentItem, int lineNumber) {
+		for (int lineNo = lineNumber; lineNo >=0; lineNo--) {
+			String tempLine = parserFileHelperUtil.getLine(textDocumentItem, lineNo).trim();
+			if (tempLine.startsWith(TO_KEY)) {
+				return "to";
+			} else if (tempLine.startsWith(FROM_KEY)) {
+				return "from";
+			} else if (tempLine.startsWith(REST_KEY)) {
+				return null;			
+			}
+		}
+		return null;
+	}
+}
--- a/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
@@ -38,11 +38,11 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 		String camelComponentURI = null;
 		String trLine = line.trim();
 		if (trLine.startsWith(URI_KEY) && URI_KEY.length() < characterPosition) {
-			camelComponentURI = trLine.substring(trLine.indexOf(URI_KEY)+ URI_KEY.length()).trim().replaceAll("\"", "");
+			camelComponentURI = extractCamelCOmponentURI(trLine, URI_KEY);
 		} else if(trLine.startsWith(FROM_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
-			camelComponentURI = trLine.substring(trLine.indexOf(FROM_KEY)+ FROM_KEY.length()).trim().replaceAll("\"", "");
+			camelComponentURI = extractCamelCOmponentURI(trLine, FROM_KEY);
 		} else if(trLine.startsWith(TO_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
-			camelComponentURI = trLine.substring(trLine.indexOf(TO_KEY)+ TO_KEY.length()).trim().replaceAll("\"", "");
+			camelComponentURI = extractCamelCOmponentURI(trLine, TO_KEY);
 		}
 		return camelComponentURI;
 	}
@@ -55,6 +55,10 @@ public class CamelKYamlDSLParser extends ParserFileHelper {
 		uriInstance.setStartPositionInDocument(new Position(position.getLine(), start));
 		uriInstance.setEndPositionInDocument(new Position(position.getLine(), start+camelComponentUri.length()));
 		return uriInstance;
+	}
+
+	private String extractCamelCOmponentURI(String trLine, String key) {
+		return trLine.substring(trLine.indexOf(key)+ key.length()).trim().replaceAll("\"", "");
 	}
 
 	private int getStartCharacterInDocumentOnLinePosition(TextDocumentItem textDocumentItem, Position position) {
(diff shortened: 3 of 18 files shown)


src/test/resources/workspace/samplewithModelineLike.yaml
```diff
@@ -0,0 +1,13 @@
+// camel-k: language=yaml
```
apupier: yaml comments are using # and not //
see https://yaml.org/spec/1.2/spec.html#id2780069
modeline is supporting both notation. See https://vim.fandom.com/wiki/Modeline_magic for several example  notations

with //, the yaml file is invalid (can be tested online on http://www.yamllint.com/ )

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,84 @@
...
+	public static final String FROM_KEY = "- from:";
+	public static final String TO_KEY = "- to:";
+
+	@Override
+	public String getCamelComponentUri(String line, int characterPosition) {
+		String camelComponentURI = null;
+		String trLine = line.trim();
+		if (trLine.startsWith(URI_KEY) && URI_KEY.length() < characterPosition) {
+			camelComponentURI = trLine.substring(trLine.indexOf(URI_KEY)+ URI_KEY.length()).trim().replaceAll("\"", "");
+		} else if(trLine.startsWith(FROM_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
```
apupier: if I understand well the expression, it expects that values are contained inside double-quotes. In fact, it can be single-quoted or having no quote at all. 
see https://yaml.org/spec/1.2/spec.html#id2786942

maybe we can support the tree relatively easily? at least the single-quote?
Otherwise, please create another issue, and the best will be to use a real yaml parser, we don't really want to bother with all these details :s

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,84 @@
...
+	public static final String REST_KEY = "rest:";
+	public static final String FROM_KEY = "- from:";
+	public static final String TO_KEY = "- to:";
+
+	@Override
+	public String getCamelComponentUri(String line, int characterPosition) {
+		String camelComponentURI = null;
+		String trLine = line.trim();
+		if (trLine.startsWith(URI_KEY) && URI_KEY.length() < characterPosition) {
+			camelComponentURI = trLine.substring(trLine.indexOf(URI_KEY)+ URI_KEY.length()).trim().replaceAll("\"", "");
```
apupier: What about extracting a function for the 3 same calls with only the KEy that is changing?

something like:
```
public void String extractCamelCOmponentURI(String trLine, String key) {
return trLine.substring(trLine.indexOf(key)+ key.length()).trim().replaceAll("\"", "")
}
```

apupier: Won't the replaceAll will also replace double-quotes inside camel URI? (need to check tomorrow)
isn't only the first and last that we want to get rid of?

lhein:
> we will have some false completion on the uri: of the rest tag. I think worth creating an issue to improve it later.

I tested it in VSCode and it works as expected.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: will check that. Thanks for the pointer.

src/test/resources/workspace/samplewithModelineLike.yaml
lhein: good catch. Will fix that.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: I'd say it pretty much depends if there is the possibility of " inside the camel uri. I will check that.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: created https://github.com/camel-tooling/camel-language-server/issues/292 to track the issue.

src/test/resources/workspace/samplewithModelineLike.yaml
lhein: regarding to the spec comments start with #<space>. Will implement it that way.

lhein:
I double checked it in VSCode and the uri beneath rest: is not considered for Camel URI completion. All looks good to me so far. Feel free to ping back if you still think its an issue.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,88 @@
...
+		return line.indexOf('"') + 1;
+	}
+
+	@Override
+	public int getPositionInCamelURI(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return position.getCharacter() - line.indexOf('"')-1;
+	}
+
+	public String getCorrespondingLine(TextDocumentItem textDocumentItem, int lineNumber) {
```
apupier: it is returning the corresponding type, not the corresponding line

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/YamlDSLModelHelper.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.cameltooling.lsp.internal.instancemodel;
+
+public class YamlDSLModelHelper implements DSLModelHelper {
+	
+	private String line;
+
+	public YamlDSLModelHelper(String line) {
```
apupier: this is the type (from/to) which is provided , not the line

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,88 @@
...
+	public CamelURIInstance createCamelURIInstance(TextDocumentItem textDocumentItem, Position position,
+			String camelComponentUri) {
+		CamelURIInstance uriInstance = new CamelURIInstance(camelComponentUri, new YamlDSLModelHelper(getCorrespondingLine(textDocumentItem, position.getLine())), textDocumentItem);
+		int start = getStartCharacterInDocumentOnLinePosition(textDocumentItem, position);
+		uriInstance.setStartPositionInDocument(new Position(position.getLine(), start));
+		uriInstance.setEndPositionInDocument(new Position(position.getLine(), start+camelComponentUri.length()));
+		return uriInstance;
+	}
+
+	private String extractCamelCOmponentURI(String trLine, String key) {
```
apupier: typo extractCamelCOmponentURI -> extractCamelComponentURI

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,88 @@
...
+
+	public String getCorrespondingLine(TextDocumentItem textDocumentItem, int lineNumber) {
+		for (int lineNo = lineNumber; lineNo >=0; lineNo--) {
+			String tempLine = parserFileHelperUtil.getLine(textDocumentItem, lineNo).trim();
+			if (tempLine.startsWith(TO_KEY)) {
+				return "to";
+			} else if (tempLine.startsWith(FROM_KEY)) {
+				return "from";
+			} else if (tempLine.startsWith(REST_KEY)) {
+				return null;
```
apupier: extra space at end of line

apupier:
> I double checked it in VSCode and the uri beneath rest: is not considered for Camel URI completion. All looks good to me so far. Feel free to ping back if you still think its an issue.

by reading the code I wasn't able to understand how it was handled. By debugging step by step I understood how ti is done. I pointed some places to rename which I think will allow to understand just by reading

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: damn, I shouldn't blindly copy/paste what other people suggest in PR reviews :D

apupier: never do that! Some suggestions can be really dumb! :-)

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: removing the tabs

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: added

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/YamlDSLModelHelper.java
lhein: fixed

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: fixed

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: fixed

lhein:
ready for a next review round @apupier

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,88 @@
...
+	public static final String TO_KEY = "- to:";
+
+	@Override
+	public String getCamelComponentUri(String line, int characterPosition) {
+		String camelComponentURI = null;
+		String trLine = line.trim();
+		if (trLine.startsWith(URI_KEY) && URI_KEY.length() < characterPosition) {
+			camelComponentURI = extractCamelComponentURI(trLine, URI_KEY);
+		} else if(trLine.startsWith(FROM_KEY) && trLine.indexOf('"') >= trLine.indexOf(':') + 1) {
+			camelComponentURI = extractCamelComponentURI(trLine, FROM_KEY);
```
apupier: now that there is the test with the from, I notice that this line is not used. In fact, This case cannot happen. For the from:, there is always the "uri:"
Removing the line will simplify code

lhein: I am not sure I can follow your thoughts. Can u elaborate a bit more how you come to this conclusion?

apupier: you added a test testign completion for the completion for a "from".
i noticed that this line of code is still not covered. So I investigated why.
In fact, for the "from", there will always be a "uri" attribute too. Otherwise the yaml route cannot be valid.

The potential shortcut that we have for the "to" is not possible because we have at least a to after the "steps" inside it.

pom.xml
```diff
@@ -305,6 +305,11 @@
 			<version>${awaitility.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.21</version>
```
apupier: the latest version is 1.25 https://mvnrepository.com/artifact/org.yaml/snakeyaml/1.25

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,112 @@
...
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		Map m = extractMapFromYaml(obj);
+		return (String)m.values().toArray()[0];
+	}
+
+	private Map extractMapFromYaml(Object o) {
+		if (o instanceof List) {
+			List l = (List)o;
+			if (l.get(0) instanceof Map) {
```
apupier: might need to check that the list is not empty

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,112 @@
...
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		return extractMapFromYaml(obj);
+	}
+
+	private String extractUriFromYamlData(String line) {
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		Map m = extractMapFromYaml(obj);
+		return (String)m.values().toArray()[0];
```
apupier: might need to check that the map is not null

apupier: might need to check that there is not an empty map

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,112 @@
...
+	private Map parseYaml(String line) {
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		return extractMapFromYaml(obj);
+	}
+
+	private String extractUriFromYamlData(String line) {
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		Map m = extractMapFromYaml(obj);
```
apupier: using genericity will avoid several warnings (even if agree that for now we can only use Map<?,?> which is not providing much)

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,112 @@
...
+	public int getPositionInCamelURI(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return position.getCharacter() - line.indexOf('"')-1;
+	}
+
+	public String getCorrespondingType(TextDocumentItem textDocumentItem, int lineNumber) {
+		for (int lineNo = lineNumber; lineNo >=0; lineNo--) {
+			String tempLine = parserFileHelperUtil.getLine(textDocumentItem, lineNo);
+			Map data = parseYaml(tempLine);
+			if (data.containsKey(TO_KEY)) {
```
apupier: migth need to check that the returned data is nto null

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,112 @@
...
+	private int getStartCharacterInDocumentOnLinePosition(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		String uri = extractUriFromYamlData(line);
+		return line.indexOf(uri);
+	}
+
+	@Override
+	public int getPositionInCamelURI(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return position.getCharacter() - line.indexOf('"')-1;
```
apupier: I think this line is the last one forcing to use double-quotes notation. Marking for reference to https://github.com/camel-tooling/camel-language-server/issues/292

pom.xml
apupier: provided https://github.com/lhein/camel-language-server/pull/5

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
apupier: created https://github.com/lhein/camel-language-server/pull/6

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
apupier: the usage of the parser is not fixing the problem of double-quotes inside it because the start character is not found correctly

i tried to call the completion at the end of the to uri
```
# camel-k: language=yaml
- id: "rest"
  group: "routes"
  rest:
    verb: "post"
    uri: "/api/route"
    accepts: "text/plain"
    binding-mode: "off"
    steps:
      - convert-body:
          type: "java.lang.String"
      - to:
          uri: "test:name?delimiter=\"&"
```

apupier: proposed a fix in this PR https://github.com/lhein/camel-language-server/pull/7

apupier:
> * the completion is not working when there is an empty uri attributes inside double-quotes

proposal of fix https://github.com/lhein/camel-language-server/pull/7

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: closing this conversation because I provided a more generic solution

pom.xml
lhein: done

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: done

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: done

lhein: done

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: done

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: fixed

lhein:
I fixed the logic for the escape chars and the empty uris. Added some new tests to also cover single quote usage and without quotes usage.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+		int startPos = line.indexOf(':')+2;
+		if (line.indexOf("'", startPos) != -1) {
+			// we use single quotes
+			return line.indexOf("'", startPos)+1;
+		} else if (line.indexOf('"', startPos) != -1) {
+			// we use double quotes
+			return line.indexOf('"', startPos)+1;
+		} else {
+			// we don't use any quotes
+			return startPos;
```
apupier: it is missing the case of quotes used inside the Camel URI with plain style.
But ok to list it as a limitation and/or report an issue to improve that later

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		Map<?, ?> m = extractMapFromYaml(obj);
+		return getNonNullValue(m.values().toArray()[0]);
+	}
+
+	private Map<?, ?> extractMapFromYaml(Object o) {
+		if (o instanceof List) {
+			List<?> l = (List<?>)o;
+			if (l.size()>0 && l.get(0) instanceof Map) {
```
apupier: more readabale to use !l.isEmpty() than l.size()>0

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+	public static final String URI_KEY = "uri";
+	public static final String REST_KEY = "rest";
+	public static final String FROM_KEY = "from";
+	public static final String TO_KEY = "to";
+
+	@Override
+	public String getCamelComponentUri(String line, int characterPosition) {
+		String camelComponentURI = null;
+		Map<?, ?> data = parseYaml(line);
+		if (data.containsKey(URI_KEY) && URI_KEY.length() < characterPosition) {
```
apupier: SonarlInt is reporting that data can be null here, would worth addign a null check

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		return extractMapFromYaml(obj);
+	}
+
+	private String extractUriFromYamlData(String line) {
+		Yaml yaml = new Yaml();
+		Object obj = yaml.load(line);
+		Map<?, ?> m = extractMapFromYaml(obj);
+		return getNonNullValue(m.values().toArray()[0]);
```
apupier: SonarlInt is reporting that data can be null here, would worth adding a null check

apupier: Shoudl we check that m.values() is not empty before retrieving the first element?

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+				res += "\\\"";
+			}
+			res += parts[i];
+		}
+		return res;
+	}
+
+	private int findStartPositionOfEmptyURI(String line) {
+		int startPos = line.indexOf(':')+2;
+		if (line.indexOf("'", startPos) != -1) {
```
apupier: better to use line.indexOf with char for faster performance as mentioned by SonarLint

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+			// we use single quotes
+			return line.indexOf("'", startPos)+1;
```
apupier: better to use line.indexOf with char for faster performance as mentioned by SonarLint

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return position.getCharacter() - findStartPositionOfEmptyURI(line);
+	}
+
+	private String repairLostEscapeChars(String line) {
+		String[] parts = line.split("\"");
+		String res = "";
+		for (int i = 0; i < parts.length; i++) {
+			if (i>0) {
+				res += "\\\"";
```
apupier: SonarLint is rcommendign to use a StringBuilder for performance reasons

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+			return findStartPositionOfEmptyURI(line);
+		}
+		uri = repairLostEscapeChars(uri);
+		return line.indexOf(uri);
+	}
+
+	@Override
+	public int getPositionInCamelURI(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		return position.getCharacter() - findStartPositionOfEmptyURI(line);
```
apupier: why itis assuming that the line contains an empty URi?

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	private String repairLostEscapeChars(String line) {
```
apupier: as you want to support all styles (double-quote/single-quote/plain style) in this PR, here you also need to escape single quotes when inside single quote.
Also, it should not replace double-quotes if single-quotes or plain style is used.

apupier: a last point but I think we are fine with it. it means that we don't support "not yet valid yaml", with missing escaped characters. (but pretty sure that users will be ok without that)

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: Its just the methodName. Actually it doesn't matter.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: Funny...SonarLint plugin in VSCode says nothing about it. Will fix it.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: I guess you meant m ?

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
apupier: please rename the method then as it is misleading

lhein: renaming it

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
lhein: Yeah, I will test that in vscode editor and see what it means for users.

lhein:
I added a couple more tests to cover your findings and fixed those problems.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,208 @@
...
+			return o.toString();
+		}
+		return "";
+	}
+
+	private int getStartCharacterInDocumentOnLinePosition(TextDocumentItem textDocumentItem, Position position) {
+		String line = parserFileHelperUtil.getLine(textDocumentItem, position.getLine());
+		String stringEncloser = getStringEncloser(line);
+		String uri = extractUriFromYamlData(line);
+		if (uri.length()==0) {
```
apupier: uri.isEmpty() is more readable

apupier: Sonarlint is reporting that uri can be null

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
apupier: SonarLint is still reporting that data can be null

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
```diff
@@ -0,0 +1,208 @@
...
+					res.append(String.format("%s%s", stringEncloser, stringEncloser));
+				}
+			}
+			res.append(parts[i]);
+		}
+		return res.toString();
+	}
+
+	private int findStartPositionOfURI(String line) {
+		int separatorPos = line.indexOf(':')+2;
```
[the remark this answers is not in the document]
lhein: Will double check this.

lhein:
The line number of your last sonar issue doesn't match reality in the file.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelKYamlDSLParser.java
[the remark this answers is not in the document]
lhein: The thing which confuses me here is that if you have no whitespace between : and the uri value then the coloring of the uri keyword changes. Wondering if it is allowed at all to have no whitespaces in between.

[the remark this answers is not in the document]
lhein: I tested that case in an online validator and it takes the non-spaces-containing key/value pair and unifies it into a single value. So I guess the whitespace is required.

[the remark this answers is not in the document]
apupier: you're right!

Issue #260, opened 2019-07-31
Support websocket connection


apupier:
Currently the connection is on stdin/stdout.
it would be useful to have websocket connection available so can be consumed from https://github.com/wylieconlon/lsp-editor-adapter

apupier:
Websocket connection in lsp4J is coming in 0.8 version, planned to be released in August https://github.com/eclipse/lsp4j/issues/203

apupier:
possible start o fimplementation https://github.com/apupier/camel-language-server-1/tree/260-provideWebsocketConnection but not sure what is the simplest way to use the javax.websocket.server.ServerEndpointConfig

apupier:
as we ar enot using a classical SpringBoot project but just repackaging using springboot. I'm wondering if it is not easier/lighter to use the reference implementation Tyrus: https://tyrus-project.github.io/documentation/1.12/index/getting-started.html

apupier:
follow-up https://github.com/camel-tooling/camel-language-server/issues/303 to allow configurable connection

apupier:
to test it manually, recommend to use the CodeMirror example https://github.com/camel-tooling/camel-lsp-client-codemirror

djelinek:
verified with GitHub codemirror example and locally started LSP server with --websocket parameter

PR #299, opened 2019-09-23
Provide websocket support #260

apupier:
- use Tyrus, the reference implementation
- connection through non-configurable port 8025
- camel server jar needs to launched with --websocket

no test, I don't know yet how to write test for it :-(


apupier:
example to test manually with CodeMirror Client https://github.com/camel-tooling/camel-lsp-client-codemirror/pull/2

PR #331, opened 2019-12-02
FUSETOOLS2-72 - support additional components

apupier:
provided as a JSON list following the Camel Catalog syntax.

# Pull Request informations

3. Green job for PR
4. Approved PR

## PR labels default process

## Tests

## PR workflow progress

2. [ ] Green **job for PR**

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,21 @@ It is possible to specify a specific version of the Camel catalog. This can be p
 }
 ```
 
+## Additional Camel components
+
+It is possible to specify a list of additional Camel components. This can be provided as Initialization Option or Workspace Configuration. The format is the same for both:
+
+```
+{
+	"camel": {
+		"extra-components" : [
+			...
+		]
+	}
+}
+
+```
+
 ## Features planned
 
 * As you type reporting of parsing and compilation errors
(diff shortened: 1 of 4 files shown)


apupier:
demo video https://drive.google.com/file/d/1QnXlh4UkTclkg1w-YLt4gGPCM7qzToxE/view?usp=sharing

lhein:
do we really need that full json blob? Can't we do it by specifying additional jars for instance?

apupier:
> do we really need that full json blob?

It is the easiest to implement and most flexible solution that can cover all cases.

> Can't we do it by specifying additional jars for instance?

it is possible but it requires more work. Next iterations can be:
- point to a list of json files
- point to a jar containing the json file
- provide a maven groupid:artifactid:version
- read in the dependencies of the project to grab the additional artefacts

lhein:
would it be possible to not be forced to open up the settings json file in an editor but to insert that blob via the preferences ui instead?

lhein:
`"type": "array",
 "default": null,
 "items": {
    "type": "string"
 }`
?

apupier:
> would it be possible to not be forced to open up the settings json file in an editor but to insert that blob via the preferences ui instead?

I have not found how to do it with VS Code API.
The advantage of the settings json editor is that it isa JSOn editor and that the parameter is a JSON objects, so benefiting from validation.

apupier:
> `"type": "array", "default": null, "items": { "type": "string" }`
> ?

they are JSON Objects, not strings. I don't think we have a JSon schema for the Camel Catalog :-(

lhein:
You are parsing a json string atm too. I don't see why that would be different.

apupier:
with just "string", the UI is even less readable:

Screenshot from 2019-12-02 16-22-46
Screenshot from 2019-12-02 16-24-18

we are missing the JSon syntax validation in the settings.json:

Screenshot from 2019-12-02 16-24-56

lhein:
ok, so lets improve that over time with better options.

PR #337, opened 2019-12-10
FUSETOOLS2-82 - all properties files are supported

apupier:
previously only application.properties were supported for camel.xxx and
all names for Kafka full Camel URI completion.

now all properties files are supported with camel.xxx, keeping the
support of full Camel URI support for specific source and sink Camel
Kafka Connect

Signed-off-by: Aurélien Pupier <apupier@redhat.com>

completionAllPropertiesFiles

# Pull Request informations

## Rebase & Merge default requirements

1. Green build for master branch
2. Wait 24 hours after PR creation
3. Green job for PR
4. Approved PR

## PR labels default process

- READY_FOR_REVIEW  &rightarrow;  REVIEW_DONE  &rightarrow;  READY_FOR_MERGE

## Tests

- [ ] Are there **Unit tests**?
- [ ] Are there **Integration tests**?
- [ ] Do we need a new **UI test**?

## PR workflow progress

1. [ ] Tagged with relevant **PR labels**
2. [ ] Green **job for PR**
3. [ ] PR was created more than **24 hours ago** or **All committers approved** it
4. [ ] Green **master** branch build

PR #338, opened 2019-12-10
Fusetools2 82 properties components with parameters completion

apupier:
requires https://github.com/camel-tooling/camel-language-server/pull/337

Screenshot from 2019-12-10 16-42-54

# Pull Request informations

3. Green job for PR
4. Approved PR

## PR labels default process

## Tests

## PR workflow progress

2. [ ] Green **job for PR**

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentOptionNamesCompletionFuture.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelComponentOptionNamesCompletionFuture.java
@@ -1,0 +1,52 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


apupier:
CamelKafka is using properties files to also define Camel URI.
So if we are inside a Camel URI in properties file, we need to delegate to the CamelEndpointCompletionProcessor. The only code that was capable to detect this condition was inside CamelKafkaConnectParserDSL

But you're right, i think I can extract a part of it.

apupier:
ok so reworked a bit to have finer grained method to detect if we need to use the completion for Camel URI or the one for Properties.

This part might be better placed in a "CamelPropertyFileValue" and delegate at this place only but as currently the model doesn't take care of Property values, I tend to think that it is simpler to let it here for this iteration.

For "move the Uri parser to a more generic utility class", re-reading the code, it seems legit to delegate the detection of a Camel Kafka specific property to the CamelKafkaConnectDSLParser.
With new version of the PR, the extracted class would contain only:
```
public boolean isInsideACamelUri(String line, int characterPosition) {
		return line.startsWith(CAMEL_SOURCE_URL) && CAMEL_SOURCE_URL.length() < characterPosition 
				|| line.startsWith(CAMEL_SINK_URL) && CAMEL_SINK_URL.length() < characterPosition;
	}
```
and some cross-references to the constants between the CamelKafkaConnectDSLParser and the utility class.
(at least, it will be the case if we keep the current restriction on notation for properties file that it must be with '=' and without spaces)

@bfitzpat With these explanations, do you agree to keep it in the CamelKafkaConnectDSLParser class? or still prefer to extract a small utility class with cross-references for the constants? (or I missed the concern that you are raising and went in the wrong direction?)

apupier:
If I understood well, it would give something like that https://github.com/apupier/camel-language-server/pull/1
I'm not convinced that it is clearer.

apupier:
> I'm ok with the kafka approach for now, but think we should extract out anything we're going to use more broadly wherever possible. But I get the difficulty

I created https://issues.redhat.com/browse/FUSETOOLS2-132 to be sure to improve this part of the code

PR #395, opened 2020-06-25
FUSETOOLS2-513 - provide completion for top level modeline options

apupier:
based on https://github.com/camel-tooling/camel-language-server/pull/393

- hardcoded list for now. The "useful" ones, as mentioned in Camel K
documentation, are not available dynamically for now anyway.
- it is providing completion only at the end of a modeline, not in the
middle

completionModelineIteration1

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelKModelineOptionNames.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/completion/CamelKModelineOptionNames.java
@@ -1,0 +1,54 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/cameltooling/lsp/internal/completion/CamelKModelineOptionNames.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.eclipse.lsp4j.CompletionItem;
+
+public class CamelKModelineOptionNames {
+	
+	private CamelKModelineOptionNames() {}
+
+	private static final List<CompletionItem> COMPLETION_ITEMS;
+	
+	static {
```
lhein, line 30: I am wondering if we can grab possible keys in a more dynamic way.

apupier: hardcoded list for now. The "useful" ones, as mentioned in Camel K
documentation, are not available dynamically for now anyway.

see https://issues.redhat.com/browse/FUSETOOLS2-516 that would allow a dynamic retrieval of catalog.

PR #397, opened 2020-06-26
Fusetools2 519 provide completion for trait property name modeline

apupier:
completionForTraitPropertyName

based on https://github.com/camel-tooling/camel-language-server/pull/396

PR #398, opened 2020-06-26
Fusetools2 520 provide completion for trait property name with default value

apupier:
based on https://github.com/camel-tooling/camel-language-server/pull/397

defaultValueAddedAfterCompletion

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/completion/modeline/TraitProperty.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/completion/modeline/TraitProperty.java
@@ -28,6 +28,13 @@ public class TraitProperty {
 	public CompletionItem createCompletionItem() {
 		CompletionItem completionItem = new CompletionItem(name);
 		completionItem.setDocumentation(description);
+		if (defaultValue != null) {
+			if("int".equals(type) && defaultValue instanceof Double) {
+				completionItem.setInsertText(name+"="+((Double)defaultValue).intValue());
+			} else {
+				completionItem.setInsertText(name+"="+defaultValue);
+			}
+		}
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/cameltooling/lsp/internal/completion/modeline/TraitProperty.java
```diff
@@ -28,6 +28,13 @@
 	public CompletionItem createCompletionItem() {
 		CompletionItem completionItem = new CompletionItem(name);
 		completionItem.setDocumentation(description);
+		if (defaultValue != null) {
+			if("int".equals(type) && defaultValue instanceof Double) {
+				completionItem.setInsertText(name+"="+((Double)defaultValue).intValue());
```
lhein, line 33: how about using String.format instead of multiple concatenations?

src/main/java/com/github/cameltooling/lsp/internal/completion/modeline/TraitProperty.java
```diff
@@ -28,6 +28,13 @@
...
+			} else {
+				completionItem.setInsertText(name+"="+defaultValue);
```
lhein, line 35: see above comment about String.format

src/main/java/com/github/cameltooling/lsp/internal/completion/modeline/TraitProperty.java
apupier: what is the advantage?

https://redfin.engineering/java-string-concatenation-which-way-is-best-8f590a7d22a8

`Avoid using String.format() when possible. It is slow and difficult to read when you have more than two variables.`

PR #401, opened 2020-06-29
Fusetools2 526 completion for camel component dependency

apupier:
based on https://github.com/camel-tooling/camel-language-server/pull/397

camelComponentDependencyCompletionInModeline

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineDependencyOption.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineDependencyOption.java
@@ -18,33 +18,22 @@ package com.github.cameltooling.lsp.internal.modelinemodel;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 import org.apache.camel.catalog.CamelCatalog;
 import org.eclipse.lsp4j.CompletionItem;
 
-import com.github.cameltooling.lsp.internal.completion.modeline.CamelKTraitManager;
+import com.github.cameltooling.lsp.internal.completion.CompletionResolverUtils;
+import com.github.cameltooling.model.util.ModelHelper;
 
-public class CamelKModelineTraitOption implements ICamelKModelineOptionValue {
+public class CamelKModelineDependencyOption implements ICamelKModelineOptionValue {
 
+	private String value;
 	private int startPosition;
-	private int endPosition;
-	private String optionValue;
-	private String traitDefinitionName;
 
-	public CamelKModelineTraitOption(String optionValue, int startPosition) {
-		this.optionValue = optionValue;
+	public CamelKModelineDependencyOption(String value, int startPosition) {
+		this.value = value;
 		this.startPosition = startPosition;
-		this.endPosition = startPosition + optionValue.length();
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineDependencyOption.java
```diff
@@ -0,0 +1,76 @@
...
+		return startPosition <= position && position <= getEndPositionInLine();
+	}
+	
+	@Override
+	public CompletableFuture<List<CompletionItem>> getCompletions(int position, CompletableFuture<CamelCatalog> camelCatalog) {
+		if(getStartPositionInLine() == position) {
+			return camelCatalog.thenApply(catalog -> catalog.findComponentNames().stream()
+				.map(componentName -> ModelHelper.generateComponentModel(catalog.componentJSonSchema(componentName), true))
+				.map(componentModel -> {
+					CompletionItem completionItem = new CompletionItem("camel-"+componentModel.getScheme());
```
lhein, line 65: so the dependency takes the scheme? or would it be better to get the dependency instead?

src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineOption.java
```diff
@@ -43,6 +44,8 @@ private ICamelKModelineOptionValue createOptionValue(String option, int nameValu
 			int startPosition = getStartPositionInLine() + optionName.length() + 1;
 			if("trait".equals(optionName)) {
 				return new CamelKModelineTraitOption(value, startPosition);
+			} else if("dependency".equals(optionName)) {
```
lhein, line 47: I would also suggest to extract "dependency" into a constant.

src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineDependencyOption.java
lhein: I think that the scheme must not fit the component name in any case.

apupier: Claus said that it is the component name https://chat.google.com/room/AAAAeFFdN0E/zrkZaoCUbAw
The component name is the scheme is in the catalog, isn't it?

apupier: hum maybe it is getLabel

apupier: or getLabel, getArtifactId, getScheme, getTitle... asked to Claus on the chat

lhein: image

lhein: was more thinking of the artifactId which would also remove the need to prepend the "camel-" string.

apupier: so, Claus said it is the artifactId. let's go with that, well spotted

PR #403, opened 2020-06-29
Fusetools2 527 completion for maven dependency

apupier:
based on https://github.com/camel-tooling/camel-language-server/pull/401

completionDependencyMavenInModeline


apupier:
> Obviously this change doesn't do any kind of validation on the maven coordinates provided. Is that a separate concern that we can cover in another PR?

That's possible; it is pulling several concerns (including performance and scope of Maven repositories to use) and use of the Maven API directly as we cannot use the m2e one which is a lot more convenient.
The advantage of this completion si at least that users is aware of the correct notation (separator sometimes slash sometimes double-dot)
I reported https://issues.redhat.com/browse/FUSETOOLS2-542 and https://issues.redhat.com/browse/FUSETOOLS2-542 to improve in the future

PR #423, opened 2020-07-16
FUSETOOLS2-560 - integrate camel-tooling-common

apupier:
benefits:
- a single repository
   - less updates of dependency to handle
   - easier synchronization of dependency update
   - single release to do
- removal of snapshot repository access
    - it will allow to remove xtend direct dependency (as it was a
workaround)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,6 @@
 		<junit.version>5.6.2</junit.version>
 		<assertj.version>3.16.1</assertj.version>
 		<camel.version>3.3.0</camel.version>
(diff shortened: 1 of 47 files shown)


apupier:
noticed that the moved code contains a lot of bad indentation. I will force-push a version with fix for that to start from a cleaner state in this repository.

apupier:
> noticed that the moved code contains a lot of bad indentation. I will force-push a version with fix for that to start from a cleaner state in this repository.

done

PR #426, opened 2020-07-20
FUSETOOLS2-504 - add support for new duration error type

apupier:
also improved test to check that there is no diagnostic with empty
message. The diagnostic in this case is not very useful and they are
ignored in VS Code UI.

Screenshot from 2020-07-20 12-38-14


apupier:
merging withotu awaiting 24 hours or everyone to review as it is to fix the master of camel-lsp-client-vscode

PR #439, opened 2020-08-21
FUSETOOLS2-608 - ensure javadoc issues detected in pr

apupier:
https://github.com/camel-tooling/camel-language-server/pull/438 to be reviewed and merged first

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,18 +14,15 @@ jobs:
     - stage: test
       jdk: openjdk8
       script:
-        - mvn install -P !build-extras -DskipTests=true -B -V
-        - mvn test -P !build-extras -B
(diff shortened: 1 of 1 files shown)


apupier:
see https://travis-ci.org/github/camel-tooling/camel-language-server/jobs/720555039#L594 , the javadoc is now generated in PR too

lhein (approved):

lhein:
so we don't need the artifact installed in the local .m2 repo? If so we are good otherwise verify isn't enough.

apupier:
> so we don't need the artifact installed in the local .m2 repo? If so we are good otherwise verify isn't enough.

no, there is no need to install in local .m2. There is a single maven reactor for the build.

PR #454, opened 2020-09-10
FUSETOOLS2-536 - provide completion for open-api in Camel K modeline

apupier:
filters on json, yaml and yml files as a first iteration. We don' tknow
yet what is the precise filter for Open APi spec v2.

Screenshot from 2020-09-10 17-13-28


bfitzpat:
If we don't know what the Open APi spec v2 holds yet -- is this premature?

apupier:
> If we don't know what the Open APi spec v2 holds yet -- is this premature?

there is for sure json/yaml files. I don't know if there are other formats and how to filter more based on content. At least, it provides completion to known potential files.

bfitzpat (approved):

PR #461, opened 2020-09-23
FUSETOOLS2-707 - filter out files in IDE config folders

apupier:
when providing completion for resource-like Camel K modeline properties.

Screenshot from 2020-09-23 10-09-51

list is currently limited to 3 IDEs and the most common folder for a Camel K project

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,17 @@ On hover, the documentation of the Camel component is available.
 On save, diagnostics on Camel URIs are updated:
 ![Diagnostic on Camel URI](./images/diagnostic.png "Diagnostic on Camel URI")
 
+### Camel K modeline support
+
+* Completion for:
+  * option names
+  * trait definition names
+  * trait property names
+  * Camel artifact id for dependency
+  * mvn dependency
+  * Camel component properties
+  * resource-like properties (`resource`, `open-api`, `property-file`). It is providing relevant sibling files filtering out some IDEs configuration folders (`.vscode`, `.settings`, `.theia`).
+
 ### WebSocket support
 
 Connection through WebSocket is supported. The server needs to be launched with `--websocket` option.
--- a/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineLocalResourceRelatedOption.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineLocalResourceRelatedOption.java
@@ -37,6 +37,10 @@ import com.github.cameltooling.lsp.internal.completion.CompletionResolverUtils;
 
 public abstract class CamelKModelineLocalResourceRelatedOption implements ICamelKModelineOptionValue {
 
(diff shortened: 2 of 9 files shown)


bfitzpat:
Would be useful to add some details on how we add more IDE specific filters or user-specific folders to the filter list

bfitzpat (changes requested):
Would be useful to have some documentation on how these files are being filtered out -- and perhaps a way for the user to see what's being filtered in case we go too far or there are additional folders/files they want to filter out.

apupier:
> Would be useful to have some documentation on how these files are being filtered out

Can you precise which documentation you have in mind please? This is a one-liner in the code. i don't get what needs to be documented.

> and perhaps a way for the user to see what's being filtered in case we go too far or there are additional folders/files they want to filter out.

I also wondered about that. I preferred not do it for these reasons:
* I think that this PR provides enough value in most cases compared to the very rare case that is filtering out and users doesn't want. In this case, the user is not blocked anyway. He/she can still provide the value by hand.
* Providing a configurable version can be done in a separate PR. And to be leveraged it will imply to modify also each clients.
* I'm not sure that it provides enough value to provide settings to configure it. It is to target very rare case and it requires to maintain the code in several clients.

another possibility that I have in mind which avoids to requires writing code for each client, is to sort with ide folders at the end instead of completely filtering them out.

bfitzpat:
> Can you precise which documentation you have in mind please? This is a one-liner in the code. i don't get what needs to be documented.

What folders are being filtered out from the IDE side? Can you include that list in the changelog at least? Or the JIRA itself, detailing which folders are being filtered out per IDE? Though the user shouldn't be using files from these folders, it would be good to at least let users know that we are now explicitly filtering them out. 

And I get the reasons for not doing a user-provided filter at this point. I would spin that off into a separate JIRA so your rationale is at least covered in case it comes up -- and then if we get enough push back on it we can investigate the individual client changes that would have to be made to support it.

apupier:
> What folders are being filtered out from the IDE side?

ok, I thought it was documentation targeting developers.

> Can you include that list in the changelog at least?

There is no changelog provided for the Camel Language Server itself. It is used mostly continuously from snapshots and usually end-users are looking to clients and not to the server part.
There is a mention specific to VS Code in the only changelog that we are currently maintaining in this [PR](https://github.com/camel-tooling/camel-lsp-client-vscode/pull/426)

> Or the JIRA itself, detailing which folders are being filtered out per IDE? Though the user shouldn't be using files from these folders, it would be good to at least let users know that we are now explicitly filtering them out.

good point.
I will add in the readme and in the JIRA.

> And I get the reasons for not doing a user-provided filter at this point. I would spin that off into a separate JIRA so your rationale is at least covered in case it comes up -- and then if we get enough push back on it we can investigate the individual client changes that would have to be made to support it.

reported https://issues.redhat.com/browse/FUSETOOLS2-715

bfitzpat (approved):

PR #463, opened 2020-09-28
Improve Contributing guide for newcomers

apupier:
- Provide information when using in Eclipse Desktop
- Include Eclipse Launch configurations in project to have them rightly
available and also mention them more easily in contributing guide

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,4 +1,8 @@
-# First Time Setup
+# First Time Setup in Eclipse Desktop
+
+Note: Eclipse Desktop IDE is providing a convenient development workflow for the Camel Language Server. That said, other IDEs can be used.
+
+## Project setup
 
 0. Fork and clone the repository
 1. Install Eclipse [Oxygen Java EE](https://www.eclipse.org/downloads/packages/release/2019-03/r/eclipse-ide-enterprise-java-developers)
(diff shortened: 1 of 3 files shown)


Contributing.md
```diff
@@ -1,4 +1,8 @@
-# First Time Setup
+# First Time Setup in Eclipse Desktop
+
+Note: Eclipse Desktop IDE is providing the most convenient development workflow for the Camel Language Server. That said, other IDEs can be used.
```
lhein, line 3: I am tempted to say this is a matter of personal bias and taste and hence does not really belong into this contributing guide except you want to definitely drive people to use Eclipse over any other IDE.

apupier: not completely a matter of taste. There are only 2 IDEs that are providing specific tooling for Language Server edition. VS Code and Eclipse Desktop. In this case, the Language Server is written in Java. Java support for Eclipse Desktop is by far better than VS Code one.
but ok, can tell that it is not "the most" but "one of the most".

lhein: I am not judging which one is better but we should simply be clear in our message. Do we want to suggest a specific IDE or do we not? If we want to suggest Eclipse, then be clear about it. That is all I wanted to state.

apupier: I want to make it clear that other IDEs can be used so that users don't think they are stuck with Eclipse.

lhein: I'd say then this is not really achieving this goal. It reads more like you should use Eclipse but there is an option to use other IDEs ;)

apupier: I wan't clear in my comments. "I want to make it clear that other IDEs can be used so that users don't think they are stuck with Eclipse." was only in opposition to your previous comment. Not for the whole intent.

More generally, in this contributing guide, I want to make it clear that Eclipse desktop is a very convenient way to develop Camel Language Server. Which is why the contributing guide is written targeting Eclipse Desktop. But it is possible to use another IDE too. Users are not stuck with Eclipse Desktop.

apupier: what about?

`Eclipse Desktop IDE is providing one of the most convenient development workflow for the Camel Language Server. This is why this guide is focused on using Eclipse Desktop IDE. Other IDEs can be used too as it remains a plain Maven Java project for the server part. The testing with the client part will require more work.`

lhein: How about `Eclipse Desktop IDE is providing a convenient development workflow for the Camel Language Server.`

lhein (approved):

Contributing.md
```diff
@@ -13,6 +17,25 @@ detect the projects and import it properly.
 3. If you discover an error on `pom.xml` after import about Tycho, you can use Quick Fix
 (Ctrl+1) to install the Tycho Maven integration.
 
+## Local debug flow with an editor
+
+Eclipse desktop allows to debug Language Server being written in the same workspace. To do that:
```
bfitzpat, line 22: Eclipse desktop allows developers to debug Language Server [code/projects?] being written in the same workspace.

Contributing.md
```diff
@@ -1,4 +1,8 @@
-# First Time Setup
+# First Time Setup in Eclipse Desktop
+
+Note: Eclipse Desktop IDE is providing a convenient development workflow for the Camel Language Server. That said, other IDEs can be used.
```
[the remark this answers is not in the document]
apupier: https://github.com/camel-tooling/camel-language-server/pull/467

Contributing.md
apupier: https://github.com/camel-tooling/camel-language-server/pull/468

PR #466, opened 2020-09-29
FUSETOOLS2-719 - include component path in outline for Java DSL

apupier:
Screenshot from 2020-09-29 14-27-32

limitation: doesn't support having several level of camel nodes on the same line. I have nerver seen a route written this way. so thought to be okay to not handle it


lhein (approved):
good addition

bfitzpat (approved):

PR #469, opened 2020-10-02
FUSETOOLS2-721 - outline of Camel endpoints for XML

apupier:
- it has kind of redundant imbrication for route/from but I preferred to
keep it as-is. It stays closer to Java outline. I think it is
interesting in outline, it allows to have different details. it is not
very nice with Ctrl+shift+O as it highlights the whole content route and
not just the from. That said, there is also the "from" highlighting only
the from line with the xml Document symbol which is usually also
installed. I do not clear advantages for one or the other option. So
picked the one easiest to maintain and better for the outline instead of
the Document symbol search which is a less visible feature.

Screenshot from 2020-10-02 10-06-45

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/documentsymbol/AbstractDocumentSymbolProcessor.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/documentsymbol/AbstractDocumentSymbolProcessor.java
@@ -1,0 +1,100 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.cameltooling.lsp.internal.documentsymbol;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.stream.Stream;
+
+import org.apache.camel.parser.model.CamelEndpointDetails;
+import org.apache.camel.parser.model.CamelNodeDetails;
+import org.eclipse.lsp4j.DocumentSymbol;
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/cameltooling/lsp/internal/documentsymbol/AbstractDocumentSymbolProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.SymbolInformation;
+import org.eclipse.lsp4j.SymbolKind;
+import org.eclipse.lsp4j.TextDocumentItem;
+import org.eclipse.lsp4j.jsonrpc.messages.Either;
+
+import com.github.cameltooling.lsp.internal.parser.ParserFileHelperUtil;
+
+public abstract class AbstractDocumentSymbolProcessor {
```
apupier, line 38: this class is mainly an extract from DocumentSymbolJavaProcessor.

src/test/java/com/github/cameltooling/lsp/internal/documentsymbol/DocumentSymbolProcessorTest.java
```diff
@@ -59,12 +64,41 @@ void testRoutesProvidedAsDocumentSymbol() throws Exception {
...
+	void testEndpointsInXml() throws Exception {
+		String textTotest =
+				"<camelContext id=\"camel\" xmlns=\"http://camel.apache.org/schema/spring\">\r\n" +
+				"    <route id=\"a route\">\r\n" +
+				"      <from uri=\"direct:cafe\"/>\r\n" +
+				"      <to uri=\"direct:drink\"/>\r\n" +
+				"    </route>\r\n" +
+				"</camelContext>\n";
+		List<Either<SymbolInformation, DocumentSymbol>> documentSymbols = testRetrieveDocumentSymbol(textTotest, 4);
+		checkSymbolInformation(documentSymbols, new Position(2, 0), new Position(3, 30), "from direct:cafe");
```
apupier, line 88: this test is where we can see that the "from" incorporates all CamelNodes. This is what allows to have the hierarchy in outline, similar to what is done in Java. The drawback is when using the "search document symbol"

end position to be replaced with (2,32) if we want to go the other way.

bfitzpat:
Question: Does this change handle a case where there are multiple routes in a single XML? See https://stackoverflow.com/questions/11192743/multiple-camel-route-processing-same-file

apupier:
> Does this change handle a case where there are multiple routes in a single XML?

yes, it is
this test https://github.com/camel-tooling/camel-language-server/pull/469/files#diff-bab2da2f6cae8e96dcf4a8bcdcce31068445ec526c117a90bb6ef3cc90936a11R154 is covering this case

visually, for instance:
Screenshot from 2020-10-22 15-07-40

bfitzpat (approved):
+1 with a green build

PR #470, opened 2020-10-06
FUSETOOLS2-735 - avoid NPE in Document Symbol when there is a space in

apupier:
path

to check:
- create folder with a space
- create a Camel Java file in it
- ensure an access oto document symbol is done (for instance by accessing outline)


bfitzpat (approved):
+1 works well for me

PR #471, opened 2020-10-07
FUSETOOLS2-755 - Provide range on Hover

apupier:
- it ensures clients to correctly updates the tooltip when several
language servers are contributing hovers on the same parts. (especially
Eclipse Desktop)
- it potentially avoids recalling the language server by the client when
the hover for the range was already provided.

on master without this PR:
wrongTooltip
with this PR:
correctlyUpdatedTooltip

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineOption.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineOption.java
@@ -23,6 +23,8 @@ import java.util.concurrent.CompletableFuture;
 import org.apache.camel.catalog.CamelCatalog;
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.Hover;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.TextDocumentItem;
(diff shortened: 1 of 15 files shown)


src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineOption.java
```diff
@@ -110,7 +112,10 @@ public boolean isInRange(int positionInLine) {
 		if(getStartPositionInLine() <= characterPosition && characterPosition <= getStartPositionInLine() + optionName.length()) {
 			String description = CamelKModelineOptionNames.getDescription(optionName);
 			if(description != null) {
-				return CompletableFuture.completedFuture(new Hover(Collections.singletonList((Either.forLeft(description)))));
+				Hover hover = new Hover();
+				hover.setContents(Collections.singletonList((Either.forLeft(description))));
+				hover.setRange(new Range(new Position(getLine(), getStartPositionInLine()), new Position(getLine(), getStartPositionInLine() + optionName.length())));
+				return CompletableFuture.completedFuture(hover);
```
apupier, line 118: special case because the model is not using a specific class for the Camel K modeline option name

bfitzpat (approved):
Definitely a better UX with the change.

PR #472, opened 2020-10-09
FUSETOOLS2-774 - provide completion for connector.class

apupier:
which is a specific Camel Kafka Connector property.
This is the first completion relying on the Camel Kafka Connector
Catalog.
Currently using a snapshot version as it is not part of a released
version

connectorClassCompletion

--> ~need to wait for Camel kafka Connector artifact to be published on Apache Snapshot repository (nightly build)~ done
--> Are we fine to use a snapshot version? or we wait fo ra release? if using snapshots, it will allow to go further more easily on other tasks related to Camel Kafka Connector. Other possibilities are to have a myriad of other PR or to create a feature branch (but needs to check how to setup the CI  for that). Or we can also wait for an official release for goign further (no date planned as far as I know)

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/propertiesfile/CamelPropertyValueInstance.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/propertiesfile/CamelPropertyValueInstance.java
@@ -26,6 +26,7 @@ import org.eclipse.lsp4j.TextDocumentItem;
 
 import com.github.cameltooling.lsp.internal.completion.CamelComponentOptionValuesCompletionsFuture;
 import com.github.cameltooling.lsp.internal.completion.CamelEndpointCompletionProcessor;
+import com.github.cameltooling.lsp.internal.completion.CamelKafkaConnectorClassCompletionProcessor;
 import com.github.cameltooling.lsp.internal.instancemodel.ILineRangeDefineable;
 import com.github.cameltooling.lsp.internal.parser.CamelKafkaUtil;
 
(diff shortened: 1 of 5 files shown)


lhein:
In general I agree that going with a snapshot version will be helpful to drive development.

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/propertiesfile/CamelPropertyValueInstance.java
```diff
@@ -48,8 +49,12 @@ public CamelPropertyValueInstance(String camelPropertyFileValue, CamelPropertyKe
 	}
 
 	public CompletableFuture<List<CompletionItem>> getCompletions(Position position, CompletableFuture<CamelCatalog> camelCatalog) {
-		if (new CamelKafkaUtil().isCamelURIForKafka(key.getCamelPropertyKey())) {
+		String propertyKey = key.getCamelPropertyKey();
+		if (new CamelKafkaUtil().isCamelURIForKafka(propertyKey)) {
 			return new CamelEndpointCompletionProcessor(textDocumentItem, camelCatalog).getCompletions(position);
+		} else if (new CamelKafkaUtil().isConnectorClassForCamelKafkaConnector(propertyKey)) {
+			String startFilter = camelPropertyValue.substring(0, position.getCharacter() - key.getEndposition() -1);
```
lhein, line 56: this logic is also used in the else case. I am wondering if we should extract it into a utils function for better sharing.

lhein (approved):
looks good to me. great to get started on the kafka connector catalog.

bfitzpat (approved):
Snapshot is probably great while we start to use some of the catalog for the kafka connector. Looks good to me!

bfitzpat (approved):

lhein (approved):

PR #478, opened 2020-10-19
FUSETOOLS2-687 - provide completion for transformers

apupier:
Screenshot from 2020-10-19 14-59-10


bfitzpat (approved):
Looks like a solid refactor for reuse purposes down the line and great incremental work on the kafka side. Good work!

PR #484, opened 2020-10-22
Fusetools2 801 provide more useful display of class names

apupier:
- use simple name in label
- use fully qualified name in details

completionWithSimpleName


bfitzpat (approved):
Good simplification and use of the tooltip/comment side

PR #487, opened 2020-10-27
FUSETOOLS2-820 - internal catalog model for new apis

apupier:
The model is handcrafted as it was before. it would be interesting to
use a JsonFactory to load everything without the need of the ModelHelper
builder for the future.


bfitzpat (approved):
Seems like a solid start until we have more support for that handcrafted JSON upstream

PR #491, opened 2020-10-29
FUSETOOLS2-827 - Fix convert to list of properties for path using / not

apupier:
as path options separator

fixConvertToList


lhein:
just wondering what happens to the aws2-s3 prefix because that is gone regarding the animated gif?

apupier:
> just wondering what happens to the aws2-s3 prefix because that is gone regarding the animated gif?

the aws2-s3 prefix is redundant for Camel kafka Connector, the connector to use is determined by the connector.class property

lhein (approved):

bfitzpat (approved):

PR #497, opened 2020-11-03
FUSETOOLS2-778 - support Camel K modeline on several lines

apupier:
- one part to cover later is to support diagnostic across the lines.
Currently, it only validates line by line
 Given the size of the PR, i think it is better to handle it in another task. No regression, it is just another new enhancement.

one example on second line:

Screenshot from 2020-11-03 14-27-23

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/CamelTextDocumentService.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/CamelTextDocumentService.java
@@ -136,7 +136,7 @@ public class CamelTextDocumentService implements TextDocumentService {
 	}
 
 	private boolean isOnCamelKModeline(int line, TextDocumentItem textDocumentItem) {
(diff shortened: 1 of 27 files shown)


bfitzpat (commented):
I'm a little confused by the test file `camelk-with-duplicated-traits-in-modeline-written-secondline.java` since the modeline is only on a single line. Is there another test with the dual dependency approach that you provided in the PR description?

apupier:
> I'm a little confused by the test file `camelk-with-duplicated-traits-in-modeline-written-secondline.java` since the modeline is only on a single line.

this file is only for the diagnostic test. it doesn't care what are on other lines (another Camel K modeline or not).  For diagnostic, it doesn't cross-check between lines as mentioned in initial description https://github.com/camel-tooling/camel-language-server/pull/497#issue-514733255

>  Is there another test with the dual dependency approach that you provided in the PR description?

given that there is no cross-line reference anywhere. I wasn't lead to that.
I will modify few tests with that. it will be closer to real-world.

bfitzpat (approved):

PR #499, opened 2020-11-04
FUSETOOLS2-817 - support new way to specify deprecation

apupier:
- continue to also use the deprecated "deprecated" field so that it is
also working on clients that have not upgraded yet.


bfitzpat (commented):
Is there an example of how this worked before vs. now? The test is very abstract and I'm looking for a more concrete use case/example?

apupier:
> Is there an example of how this worked before vs. now? The test is very abstract and I'm looking for a more concrete use case/example?

There is no visual differences in VS Code or Eclipse. It is only that the API is changing. During the transition phase, some clients will drop previous way of doing it and adopt new API at different pace. For now, I think that all editors are supporting the old API and some are supporting both API (or some are supporting none but it won't make differences anyway)

bfitzpat (approved):

PR #502, opened 2020-11-04
Fusetools2 835 hover for camel url in properties file

apupier:
requires 
hoverForcamelUrlInPropertiesFile


bfitzpat (approved):
Sweet! Loving having extra documentation that accessible. Is it plaintext? Or can it be Markdown flavored for extra formatting?

apupier:
> Is it plaintext? Or can it be Markdown flavored for extra formatting?

I have not touched this part. It is reusing the same content for hover than for Camel URI hover everywhere else. It is currently plaintext given that the catalog is providing plain text. https://github.com/camel-tooling/camel-language-server/blob/dd6db26f2664d527ff8aa247c4cc593b26535473/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/ILineRangeDefineable.java#L34 is the most used way to create the Hover content
it is possible to use markdown (if clients support it but VS Code and Eclipse Desktop are supporting it)

PR #506, opened 2020-11-12
FUSETOOLS2-866 - First Circle CI iteration: simple build with test for

apupier:
matrix

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,32 @@
+version: 2.1
+
+jobs:
+  jdk8:
+    docker:
+      - image: circleci/openjdk:8-stretch
+    steps:
+      - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+  jdk11:
+    docker:
+      - image: circleci/openjdk:11-stretch
+    steps:
+      - checkout
+      - run: mvn verify -V
+
+  jdk15:
+    docker:
+      - image: circleci/openjdk:15-buster
```
apupier, line 20: they do not provide the `stretch` version for JDK 15, so using `buster`

`stretch` is th eone used in all Java examples.

bfitzpat (approved):

PR #511, opened 2020-11-18
FUSETOOLS2-866 - remove deploy from travis

apupier:
the configuration for deployment between Circle CI and Travis are
different and incompatible. need to choose one to do it.

if wee need to switch back, this commit can be reverted

I do not see easy solution to keep both working at same time.
Would need:
- specific Maven profile activated or not depending on the CI
- different script called for the crypting key (this one sounds easier)

I do no think that it is worthy to maintain both. It would publish a build based on the same commit two times anyway keeping them both.
But if disagree, can work to provide both working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,8 @@ jobs:
       jdk: openjdk15
       script:
         - mvn verify -V
-    - stage: deploy
-      name: "Sonar Analysis and Deployment"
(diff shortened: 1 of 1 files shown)


lhein (approved):
I think one is enough.

bfitzpat (approved):
Agreed

PR #528, opened 2020-12-07
FUSETOOLS2-815 - provide validation for invalid camel.(source|sink)

apupier:
properties keys
diagnosticCamlkafkaConnectorInvalidproperty

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/diagnostic/CamelKafkaConnectorDiagnosticService.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/diagnostic/CamelKafkaConnectorDiagnosticService.java
@@ -1,0 +1,68 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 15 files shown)


bfitzpat (approved):

src/main/java/com/github/cameltooling/lsp/internal/diagnostic/CamelKafkaConnectorDiagnosticService.java
```diff
@@ -0,0 +1,68 @@
...
+
+public class CamelKafkaConnectorDiagnosticService extends DiagnosticService {
+
+	protected CamelKafkaConnectorDiagnosticService(CamelKafkaConnectorCatalogManager camelKafkaConnectorCatalogManager) {
+		super(null, camelKafkaConnectorCatalogManager);
+	}
+
+	public Collection<Diagnostic> compute(String camelText, TextDocumentItem documentItem) {
+		String docUri = documentItem.getUri();
+		if(docUri.endsWith(".properties")) {
```
lhein, line 45: Is it safe to assume the extension is always lowercase? If not I would suggest to convert to lowercase before the check.

lhein (approved):

src/main/java/com/github/cameltooling/lsp/internal/diagnostic/CamelKafkaConnectorDiagnosticService.java
apupier: I think it is safe. I never seen the properties extension written differently.

PR #532, opened 2020-12-09
FUSETOOLS2-912 - diagnostic error when mix of url and normal attributes

apupier:
mixed

diagnosticMixedNotation


bfitzpat (approved):
Nice to see some deeper validation

PR #538, opened 2020-12-11
FUSETOOLS2-908 - diagnostic for missing required properties which

apupier:
doesn't have default values


apupier:
more complex stuff needs to be impletmented in catalog to avoid false positive https://issues.redhat.com/browse/FUSETOOLS-3436?focusedCommentId=15836827&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15836827

PR #541, opened 2020-12-14
FUSETOOLS2-914 - quickfix for typo atributes

apupier:
quickfixCamelCaseCamelkafkaConnectorunknownProperties

it is suggesting only camel cased properties. Suggesting dashed case properties would be nice to be done in another iteration


lhein (approved):
nice, great addition

bfitzpat (approved):
Nice! Very useful from a development standpoint

PR #556, opened 2021-02-02
FUSETOOLS2-961 - Provide a setting to set a Runtime Provider

apupier:


bfitzpat (approved):
Nice to have an additional level of configuration -- I'm sure that will be useful as we continue to expand the functionality going forward

PR #569, opened 2021-03-17
FUSETOOLS2-1028 - replace CI testing of JDK 15 by JDK 16

apupier:


lhein:
Any reason why you dropped 15 in favor to 16 instead of having both?

apupier:
Java 15 is not an LTS.

Keeping the policy of testing:
* LTS
* latest release

lhein (approved):

bfitzpat (approved):

PR #582, opened 2021-04-12
FUSETOOLS2-1057 - provide completion for kafka topic

apupier:
first iteration:
- does not support multi-topics (which is possible for consumer
configuration)
- does not support connection that requires authentication
- use localhost:9092 by default
- use system property to override connection url (not convenient for
language server clients but useful for testing)

next iterations:
- connection url provided with a setting https://issues.redhat.com/browse/FUSETOOLS2-1059
- support credentials https://issues.redhat.com/browse/FUSETOOLS2-1060
- support of multi-topics https://issues.redhat.com/browse/FUSETOOLS2-1061
- use `brokers` attribute if provided in Camel URL property https://issues.redhat.com/browse/FUSETOOLS2-1062
- completion for some attributes that can be retrieved from a connected
instance? partitionKey? others?

https://user-images.githubusercontent.com/1105127/114393561-6f545a00-9b9a-11eb-8354-47a84ca0fd45.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,10 @@ There is document symbol for Camel Contexts and routes, on XML and Java DSL. In 
 ![Go to Symbol](./images/goToSymbol.gif "Go To Symbol")
 ![Breadcrumb](./images/breadCrumbXml.gif "Breadcrumb")
 
+## Connected mode completion for Kafka component
+
+When using the Camel Kafka component, the list of topics is dynamically retrieved from the Kafka Broker when available. By default, it tries to connect to `localhost:9092`. To connect to another instance, launch the language server with the system property `CAMEL_LANGUAGE_SERVER_KAFKA_CONNECTION_URL`. For instance, `java -DCAMEL_LANGUAGE_SERVER_KAFKA_CONNECTION_URL=localhost:9093 -jar camel-lsp-server.jar`.
+
 ## Features planned
 
 * As you type reporting of parsing and compilation errors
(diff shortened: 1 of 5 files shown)


lhein (approved):
This is a great start Aurelien!

README.md
```diff
@@ -124,6 +124,10 @@ There is document symbol for Camel Contexts and routes, on XML and Java DSL. In
 ![Go to Symbol](./images/goToSymbol.gif "Go To Symbol")
 ![Breadcrumb](./images/breadCrumbXml.gif "Breadcrumb")
 
+## Connected mode completion for Kafka component
+
+When using the Camel Kafka component, the list of topics is dynamically retrieved from the Kafka Broker when available. By default, it tries to connect to `localhost:9092`. it can be connected to another instance using the system property `CAMEL_LANGUAGE_SERVER_KAFKA_CONNECTION_URL`.
```
bfitzpat, line 129: This seems a little confusing to me - "it can be connected to another instance using the system property" - may be slightly clearer if you say `By default, it tries to connect to `localhost:9092`. To connect to another instance, update the system property `CAMEL_LANGUAGE_SERVER_KAFKA_CONNECTION_URL`.`? Maybe provide an example of using that property somehow?

bfitzpat (changes requested):
Great first step on this - my only concern is with the readme. Hoping we can add a bit more clarification on how to change the host

bfitzpat (approved):
I think that's much clearer. Thanks for the change!

PR #590, opened 2021-05-04
FUSETOOLS2-1097 - provide completion for kamelet id

apupier:
first iteration:
- ~requires locally built version until version is published on Apache
Snapshots. the version is currently main-SNAPSHOT, this will need to be updated.~ https://repository.apache.org/content/repositories/snapshots/org/apache/camel/kamelets/camel-kamelets-catalog/main-SNAPSHOT/
- no distinction between source and sink
- no description
- Catalog is reloaded on each completion call in a kamelet template Id
part

Screenshot from 2021-05-04 09-30-56

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/PathParamURIInstance.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/instancemodel/PathParamURIInstance.java
@@ -41,6 +41,7 @@ import com.github.cameltooling.lsp.internal.completion.CamelComponentSchemesComp
 import com.github.cameltooling.lsp.internal.completion.CompletionResolverUtils;
 import com.github.cameltooling.lsp.internal.completion.FilterPredicateUtils;
 import com.github.cameltooling.lsp.internal.completion.KafkaTopicCompletionProvider;
+import com.github.cameltooling.lsp.internal.completion.KameletTemplateIdCompletionProvider;
 import com.github.cameltooling.lsp.internal.settings.SettingsManager;
 
 /**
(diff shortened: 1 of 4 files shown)


apupier:
no release of kamelet catalog planned soon. i think we can go with snpashot waiting for it (will most probably happen after Productized Camel K)

lhein (approved):

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/PathParamURIInstance.java
```diff
@@ -69,8 +70,11 @@ public String getValue() {
 	@Override
 	public CompletableFuture<List<CompletionItem>> getCompletions(CompletableFuture<CamelCatalog> camelCatalog, int positionInCamelUri, TextDocumentItem docItem, SettingsManager settingsManager) {
 		if(pathParamIndex == 0) {
-			if("kafka".equals(uriInstance.getComponentName())) {
+			String componentName = uriInstance.getComponentName();
+			if("kafka".equals(componentName)) {
```
bfitzpat, line 74: Wondering if "kafka" and "kamelet" should be constants somewhere?

bfitzpat (approved):
One minor question, but looks like a good first step!

src/main/java/com/github/cameltooling/lsp/internal/instancemodel/PathParamURIInstance.java
apupier: I was also wondering it :-)

not sure where to put them:
- in this class, as currently used only here (but they will surely be used elsewhere soon)
- in a new specific class ComponentNames
- Each in a specific kafka/Kamelet class. a new one or reusing a "CompletionProvider" one

PR #605, opened 2021-05-20
FUSETOOLS2-903 - try avoid flakiness

apupier:
- provide more precise assertions
- provide more precise teardown in case previous test was affecting the
other ones, like the shutdown was not finished yet and still occupy some
ports.


apupier:
10*3 jobs in success, if not fixed, at least it shoudl nto be less stable and there will be a slightly better error if it occurs again

lhein (approved):

bfitzpat (approved):

PR #618, opened 2021-06-18
Replace lambda with a method reference

apupier:
see https://sonarcloud.io/project/issues?id=camel-lsp-server&issues=AXoZ9O-dTKya_KffcyJR&open=AXoZ9O-dTKya_KffcyJR


bfitzpat (approved):
I'm giving this a +1 with a caveat. 

Though I understand the Lambda reference in this case, I'm not sure that it's any more readable than the original way you map the value. Cooler maybe. But not more readable. I would caution against using newer, cooler technologies (and yes, I know it was introduced in Java 8, so this isn't exactly "new") that make things harder to maintain in the future.

apupier:
I think we are in the case it is commonly agreed to be more compact and readable as mentioned in Sonar rule description https://sonarcloud.io/organizations/camel-tooling/rules?open=java%3AS1612&rule_key=java%3AS1612

bfitzpat:
> I think we are in the case it is commonly agreed to be more compact and readable as mentioned in Sonar rule description https://sonarcloud.io/organizations/camel-tooling/rules?open=java%3AS1612&rule_key=java%3AS1612

Guess I'm just old then. :)

lhein (approved):

PR #626, opened 2021-06-28
FUSETOOLS2-1183 - quickfix convertproperty file

apupier:
convertActionPropertyFile

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/codeactions/ConvertCamelKPropertyFileModelineRefactorAction.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/codeactions/ConvertCamelKPropertyFileModelineRefactorAction.java
@@ -1,0 +1,82 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/cameltooling/lsp/internal/codeactions/ConvertCamelKPropertyFileModelineRefactorAction.java
```diff
@@ -0,0 +1,82 @@
...
+import com.github.cameltooling.lsp.internal.completion.modeline.CamelKModelineOptionNames;
+import com.github.cameltooling.lsp.internal.modelinemodel.CamelKModeline;
+import com.github.cameltooling.lsp.internal.modelinemodel.CamelKModelineOption;
+import com.github.cameltooling.lsp.internal.modelinemodel.ICamelKModelineOptionValue;
+import com.github.cameltooling.lsp.internal.parser.CamelKModelineParser;
+import com.github.cameltooling.lsp.internal.parser.ParserFileHelperUtil;
+
+public class ConvertCamelKPropertyFileModelineRefactorAction {
+	
+	public static final String CODE_ACTION_TITLE_CONVERT_PROPERTY_FILE = "Convert deprecated modeline options property-file to property=file: notation";
```
apupier, line 45: I'm balanced if I need to put more details here, like between which and which version it is useful but I felt it was creating a too long and not enough readable string in UI.
The quick action appears anyway only when there is the deprecated notation used.

lhein: looks good to me

lhein (approved):

PR #628, opened 2021-06-28
FUSETOOLS2-1180 - include file: as completion for value of property

apupier:
Camel K modeline option

Screenshot from 2021-06-28 16-00-25

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelinePropertyOption.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelinePropertyOption.java
@@ -16,8 +16,11 @@
  */
 package com.github.cameltooling.lsp.internal.modelinemodel;
 
+import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


lhein:
Just on first sight of the attached screenshot I am wondering why we have `camel` and then we got on the other side `file:`. Why do we complete with a `:` in that case while we do not for the other?

apupier:
there is only `camel` because it is the completion for camel component properties. it is reusing completion from properties file where it is a no-go to have the full list of available properties available from a UX perspective. From a performance perspective, it can also be a problem. This is a part that is working this way for a long time.

I was wondering if I should put the `:` visible or not, or only insert it. I decided to put the `:` after the `file` because I was finding it a bit clearer as it is a relatively special notation. But maybe better to keep only the `file` in the proposals,  improve description with an example and insert with text with `file:`

lhein:
I was just wondering. With your explanation it makes more sense now.

lhein (approved):

PR #635, opened 2021-07-08
FUSETOOLS2-1166 - factorize creation of camel route used in test

apupier:
factorized only the part with single uri with a from for blueprint and
spring in xml (which is the most used)


lhein (approved):

bfitzpat (approved):
Nice!

PR #637, opened 2021-07-16
FUSETOOLS2-1188 - provide metrics

apupier:
Telemetry is part of the LSP specification: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#telemetry_event

the event is created on the [langauge server side](https://github.com/eclipse/lemminx/tree/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/telemetry) and the [client is sending the metric](https://github.com/redhat-developer/vscode-xml/commit/ee25f3abe24015dd11e01a8e9a6dfaa4f3875af1#diff-933541f88e08dbbc0a55ba6ccddf49d2e6a53b0f4f5a71a42a6ec691d1be24a3R41-R43) to Red Hat woopra if telemetry is enabled.

in combination with https://github.com/camel-tooling/camel-lsp-client-vscode/pull/690 , we can have this kind of report: https://app.woopra.com/project/test.vscode/analyze/report/trends/c0afwq5qlb

### DIFF
--- a/USAGE_DATA.md
+++ b/USAGE_DATA.md
@@ -1,0 +1,15 @@
+# Telemetry data collection
+
+Telemetry events can be collected by the LSP client program.
+
+The following information is emitted when the language server starts:
+
+ * JVM information:
+    * Whether it is being run with Java or as a GraalVM native image (binary)
+    * The name of the vm (`java.vm.name`)
+    * The name of the runtime (`java.runtime.name`)
+    * The version of the JVM (`java.version`)
+    * The free, total, and max VM memory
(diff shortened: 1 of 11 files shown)


USAGE_DATA.md
```diff
@@ -0,0 +1,15 @@
+# Telemetry data collection
+
+Telemetry events can be collected by the LSP client program.
+
+The following information is emitted when the language server starts:
+
+ * JVM information:
+    * Whether it is being run with Java or as a GraalVM native image (binary)
```
apupier, line 8: in our case, currently always return false but it allows to have teh same structural metric data than xml ls

bfitzpat (approved):

PR #638, opened 2021-07-20
FUSETOOLS2-1222 - provide language of Camel files opened

apupier:
heuristic to detect if it s a Camel file or not is not perfect but it
the best that we were using so far.

it will allow to have this kind of report https://app.woopra.com/project/test.vscode/analyze/report/trends/j6rpgiwdyn when used in combination with https://github.com/camel-tooling/camel-lsp-client-vscode/pull/693

### DIFF
--- a/src/test/java/com/github/cameltooling/lsp/internal/telemetry/TelemetryTest.java
+++ b/src/test/java/com/github/cameltooling/lsp/internal/telemetry/TelemetryTest.java
@@ -17,12 +17,22 @@
 package com.github.cameltooling.lsp.internal.telemetry;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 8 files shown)


lhein (approved):
looking forward to see all the statistics in the coming weeks

bfitzpat (approved):

src/test/java/com/github/cameltooling/lsp/internal/telemetry/TelemetryTest.java
```diff
@@ -36,4 +42,26 @@ void testInitializationWithJvmInfoMetric() throws Exception {
...
+		TelemetryEvent telemetryEvent = telemetryEvents.get(1);
+		assertThat(telemetryEvent.name).isEqualTo(TelemetryManager.OPENED_DOCUMENT);
+		Map<String, Object> properties = telemetryEvent.properties;
+		assertThat((String)properties.get("language")).isEqualTo(extension);
+	}
+	
+	private static Stream<Arguments> testOpenedDocumentMetric() {
+		return Stream.of(
+				Arguments.of(RouteTextBuilder.createXMLSpringRoute(""), "xml"),
+				Arguments.of("camel", "java"),
```
apupier, line 60: I thought again about it. This heuristic migth be not enough restrictive for Java file. it will send an event for all Java files that have camel in one of import. To be comparable with other files, it should be more restrictive as others are more precise on (almost) only Route files. Maybe to check that it also contains a `from`.

lhein: might be better

PR #645, opened 2021-07-28
FUSETOOLS2-1231 - provide folding range for routes written in Java

apupier:
foldingRangeForJavaRoute


src/main/java/com/github/cameltooling/lsp/internal/catalog/util/CamelNodeDetailsUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.camel.parser.model.CamelNodeDetails;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextDocumentItem;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.github.cameltooling.lsp.internal.parser.ParserFileHelperUtil;
+
+public class CamelNodeDetailsUtils {
```
apupier, line 32: this class is an extract from existign code to allow factorization

lhein (approved):
great addition

PR #696, opened 2021-12-15
Upgrade log4j implementation for slf4j from 1.x to 2.16

apupier:
tricky part was to update the tests. the Appender is a bit trickier to
setup and also requires to set the log level

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,6 +46,7 @@
 		<jacoco.version>0.8.7</jacoco.version>
 		<lsp4j.version>0.12.0</lsp4j.version>
 		<slf4j.version>1.7.32</slf4j.version>
+		<log4j-slf4j.version>2.16.0</log4j-slf4j.version>
 		<junit.version>5.8.2</junit.version>
 		<assertj.version>3.21.0</assertj.version>
 		<jgit.version>6.0.0.202111291000-r</jgit.version>
@@ -240,9 +241,9 @@
 			<version>${slf4j.version}</version>
 		</dependency>
 		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
-			<version>${slf4j.version}</version>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-slf4j-impl</artifactId>
+			<version>${log4j-slf4j.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.apache.camel</groupId>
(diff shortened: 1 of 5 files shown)


src/test/java/com/github/cameltooling/lsp/internal/util/TestLoggerUtil.java
```diff
@@ -0,0 +1,47 @@
...
+		return appender;
+	}
+
+	private void updateLoggers(LoggerContext context, final Appender appender, final Configuration config) {
+		final Level level = Level.ALL;
+		for (final LoggerConfig loggerConfig : config.getLoggers().values()) {
+			loggerConfig.addAppender(appender, level, null);
+		}
+		config.getRootLogger().addAppender(appender, level, null);
+		context.updateLoggers(config);
```
apupier, line 44: without it, playing only the test class is working but inside the test suite it isn't

src/test/java/com/github/cameltooling/lsp/internal/util/TestLoggerUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
+
+public class TestLoggerUtil {
+	
+	public TestLogAppender setupLogAppender(String name) {
+		final LoggerContext context = LoggerContext.getContext(false);
+		final Configuration config = context.getConfiguration();
+		config.getRootLogger().setLevel(Level.ALL);
```
apupier, line 30: level needs to specifically set otherwise it defaults to Error level (and nto following what is defined in log configuration file)

pom.xml
```diff
@@ -240,9 +241,9 @@
 			<version>${slf4j.version}</version>
 		</dependency>
 		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
-			<version>${slf4j.version}</version>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-slf4j-impl</artifactId>
```
apupier, line 245: org.slf4j:slf4j-log4j12 seems almost dead. Activity is now on log4j-slf4j-impl

lhein: with 2.16 we are on the safe side, right?

src/test/java/com/github/cameltooling/lsp/internal/util/TestLoggerUtil.java
lhein: good catch

pom.xml
apupier: it is the latest release as of today which is not impacted by log4jshell CVE. (this version has even completely removed the feature of jndi lookup)

lhein (approved):

PR #709, opened 2022-01-04
FUSETOOLS2-1377 - Provide hint for connected mode completion for

apupier:
kubernetes and knative

choose to not check if there is a connection active or no for 2 reasons:
- avoids to potentially slow down validation
- still remains a bit interesting to have the hint when there is the connection to encourage users to use completion

limitation: available for XML and Java only

good visibility in Eclipse:
hintConnectedMode

Not good visibility in VS Code (but well that the client choice
hintConnectedModeVSCode
)


lhein (approved):
nice addition

PR #784, opened 2022-07-18
FUSETOOLS2-1677 - improve support for productized catalog

apupier:
Only the classic catalog is covered. It allows to cover the Camel
Quarkus use case. It will cover the SpringBoot based use case when it
will be productized.


joshiraez (approved):
I don't see anything outstanding on the changes

PR #793, opened 2022-07-19
Avoid direct access to attribute

apupier:
it also avoids having an unused method


joshiraez (approved):

joshiraez:
Good spot. I modified some other files but this one flew over my head. Thanks

PR #854, opened 2022-11-09
FUSETOOLS2-1870 - avoid flakiness by checking timeout more precisely

apupier:
The goal of the test is to ensure that the end-user receive the completion feedback in less than a second. The timeout on the whole test before was impacted by the time to setup the environment for the test. Keeping a timeout at test level so that in case it hangs, it does not hangs for several minutes.

### DIFF
--- a/src/test/java/com/github/cameltooling/lsp/internal/completion/KafkaTopicCompletionTest.java
+++ b/src/test/java/com/github/cameltooling/lsp/internal/completion/KafkaTopicCompletionTest.java
@@ -24,6 +24,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.DidChangeConfigurationParams;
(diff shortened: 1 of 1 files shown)


src/test/java/com/github/cameltooling/lsp/internal/completion/KafkaTopicCompletionTest.java
```diff
@@ -139,9 +141,9 @@ private void initKafkaWithTopics(String topics) {
 		System.setProperty(KafkaTopicCompletionProvider.CAMEL_LANGUAGE_SERVER_KAFKA_CONNECTION_URL, connectionURL);
 	}
 
-	private List<CompletionItem> retrieveCompletionForKafkaTopicPosition() throws URISyntaxException, InterruptedException, ExecutionException {
+	private List<CompletionItem> retrieveCompletionForKafkaTopicPosition() throws URISyntaxException, InterruptedException, ExecutionException, TimeoutException {
 		CamelLanguageServer languageServer = initLanguageServer();
-		return getCompletionFor(languageServer, new Position(0, 17)).get().getLeft();
+		return getCompletionFor(languageServer, new Position(0, 17)).get(1, TimeUnit.SECONDS).getLeft();
```
joshiraez, line 146: I think this part only puts a Timeout. Are you saying that the language server will try to avoid the timeout ensuring that it'll get the whatever is there in less than a second?

apupier: The idea is that we do not want to have the completion call taking more than a second. From a user point of view when taking more than one second it is breaking the flow.  When creating a connection to a kafka cluster, in some cases, the connection hangs. I would like to ensure that when the connection hangs, the server is https://github.com/KaotoIO/vscode-kaoto/actions/runs/3436051234/jobs/5729130030#step:9:38 responding in less than a second and returning the "static" completion and not await for a Kafka cluster to answer.

joshiraez: I see

joshiraez (approved):

PR #926, opened 2023-05-04
Deprecate Camel Kafka Connector related classes

apupier:
there was no reference in readme and there is no changelog so no other places to modify


djelinek:
maybe also mention in readme file in section about kafka?

apupier:
> maybe also mention in readme file in section about kafka?

I guess you mean the section entitled `Connected mode completion for Kafka component`. if yes, it is a different features.
Camel Kafka Connector was a specific DSL to describe some kind of Camel routes. The Connected mode for kafka component is for Camel URI completion and validation of Kafka component.

PR #947, opened 2023-06-20
FUSETOOLS2-2120 - Provide Camel Quarkus debug profile in completion

apupier:
https://github.com/camel-tooling/camel-language-server/assets/1105127/e4b31b42-6c9c-4b60-a142-1c24cf0ac01f

ideas for other iterations:
- provide completion only inside a profiles block
- provide completion only when it is a Camel Quarkus project
- include profiles block when not available


apupier:
> worthy to mention in changelog?

Yes, I need to create the PR in VS Code Camel LSP. Thanks for the reminder

PR #960, opened 2023-07-05
FUSETOOLS2-2133 - use Camel 4

apupier:
- Java 17 is required
- The `org.apache.camel.ExchangePattern` has removed `InOptionalOut`.
- Direct VM and VM component has been removed
- No Camel Kafka Connector catalog is provided, keep 3.21.0 for it but
not sure how long it would be possible

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -29,7 +29,7 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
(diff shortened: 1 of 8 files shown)


apupier:
> what the changes mean for these features? will them still be available?
> 
> * Find references for direct and direct VM components in all open Camel files.
> * Auto-completion for referenced IDs of direct, direct VM

they will still be available so that users on Camel 3 still have access to it.
For Camel 4 users, it will be available but well they will also have a validation error mentioning that the component is unknown

PR #973, opened 2023-09-06
Upgrade to Camel K 2

apupier:
It is impacting mostly the Traits completion/validation

* Embed the Camel K CRDs for now. For next iteration, it would nice to:
use one from Camel K CRD jar (or another upstream artifact) and then
also allow possibility to load from the cluster
* Requires to use Jackson Dataformat loader as Snake Yaml one is less
permissive and throwing errors
* since kubernetes client 6.2.0, no more need of a specific mock
dependency https://github.com/fabric8io/kubernetes-client/issues/3924
* No more providing default values as they not yet in CRD
https://github.com/apache/camel-k/issues/4726
* Some description are now a lot less precise, adapted tests and
reported https://github.com/apache/camel-k/issues/4727


pom.xml
```diff
@@ -366,4 +360,10 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
+	<repositories>
+		<repository>
+			<id>staging camel kamelets 4.0.0</id>
+			<url>https://repository.apache.org/content/repositories/orgapachecamel-1607</url>
+		</repository>
+	</repositories>
```
apupier, line 368: to be removed when Camel Kamelet catalog 4.0.0 release is approved

djelinek (approved):

PR #990, opened 2023-10-02
fix FUSETOOLS2-2178: Provide completion for secrets coming from Kubernetes Context

Delawen:
It searches in all namespaces if there is a kubernetes connection configured.
It doesn't discriminate by name of the property, but detects when a {{secret: is being written as value.

Maybe we could improve this, so when the value of a property starts with `{{'' it shows suggestions for different placeholder types, and once you select the type of placeholder (`secret`, `configmap`,...), it expands the auto-completions to specific values.

## Tests

- [ ] Are there **Unit tests**?
- [ ] Are there **Integration tests**?
- [ ] Do we need a new **UI test**?

## PR workflow progress

1. [ ] Tagged with relevant **PR labels**
2. [ ] Green **job for PR**
3. [ ] PR was created more than **24 hours ago** or **All committers approved** it
4. [ ] Green **main** branch build

PR #997, opened 2023-10-13
feature: Auto-completion for kubernetes mid-string

Delawen:
Continue #990 and #995

Now instead of auto-completing the whole value of the parameter, the string can contain several placeholders and other pieces of string that won't get removed by the auto-completion.

[Screencast from 2023-10-13 13-49-40.webm](https://github.com/camel-tooling/camel-language-server/assets/726590/60b33a05-ee2d-4701-8fab-9d2e35ac7a2e)


Delawen:
Not sure if this deserves a changelog or it can be included as part of the kubernetes context auto-completion lines in the changelog.

apupier:
> Not sure if this deserves a changelog or it can be included as part of the kubernetes context auto-completion lines in the changelog.

given that there was no release with it, i think we cna consider it as included in the other parts

apupier:
Please rebase so that the PR is launched on Java 21.

(and it is a good habit to always rebase when providing new changes to ensure it is not diverging too much from main branch)

apupier (approved):

PR #1001, opened 2023-10-18
Part of FUSETOOLS2-1579: Add simple language autocompletion (basic)

Delawen:
Add some simple language placeholders that can be useful. It does not check the context or the route to extract "dynamic" values like headers or exchangeProperties.

The values provided with this auto-completion are:

`${camelId}`, `${exchange}`, `${exchangeId}`, `${id}`,`${messageTimestamp}`, `${body}`, `${bodyOneLine}`, `${prettyBody}`, `${headers}`, `${exception}`, `${exception.message}`, `${exception.stacktrace}`, `${routeId}`, `${stepId}`, `${threadId}`, `${threadName}`, `${hostname}`, `${null}`, `${messageHistory}`

It also refactor common parts of the code with the kubernetes auto-completion of secrets and config maps.

[Screencast from 2023-10-18 10-47-41.webm](https://github.com/camel-tooling/camel-language-server/assets/726590/d5141f10-df6c-485e-be05-d509f3785446)

PR #1027, opened 2023-11-27
feature: Support multiline in YAML camel uri autocompletion

Delawen:
This refactor is able to handle multi-line selection in YAML. 

[Screencast from 2023-11-27 15-47-17.webm](https://github.com/camel-tooling/camel-language-server/assets/726590/c5db6f44-3053-4ecd-af70-14e118a028eb)

This is for preparation for https://github.com/camel-tooling/camel-language-server/pull/1001 which will require multiline. Still needs more features in-between before merging #1001 because we need to be able to give more context to camel uris. But now at least you can process whole YAML elements that are on different lines.

### DIFF
--- a/src/main/java/com/github/cameltooling/lsp/internal/hover/CamelKModelineHoverProcessor.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/hover/CamelKModelineHoverProcessor.java
@@ -36,7 +36,8 @@ public class CamelKModelineHoverProcessor {
 
 	public CompletableFuture<Hover> getHover(Position position, CompletableFuture<CamelCatalog> camelCatalog) {
 		int lineNumber = position.getLine();
-		CamelKModeline camelKModeline = new CamelKModeline(new ParserFileHelperUtil().getLine(textDocumentItem, lineNumber), textDocumentItem, lineNumber);
+		CamelKModeline camelKModeline = new CamelKModeline(new ParserFileHelperUtil().getLines(textDocumentItem,
+				lineNumber, lineNumber + 1), textDocumentItem, lineNumber, lineNumber);
 		return camelKModeline.getHover(position, camelCatalog);
 	}
 
--- a/src/main/java/com/github/cameltooling/lsp/internal/hover/CamelKModelineHoverProcessor.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/hover/CamelKModelineHoverProcessor.java
@@ -37,7 +37,7 @@ public class CamelKModelineHoverProcessor {
 	public CompletableFuture<Hover> getHover(Position position, CompletableFuture<CamelCatalog> camelCatalog) {
 		int lineNumber = position.getLine();
 		CamelKModeline camelKModeline = new CamelKModeline(new ParserFileHelperUtil().getLines(textDocumentItem,
-				lineNumber, lineNumber + 1), textDocumentItem, lineNumber, lineNumber);
+				lineNumber, lineNumber), textDocumentItem, lineNumber, lineNumber);
 		return camelKModeline.getHover(position, camelCatalog);
 	}
 
--- a/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
@@ -264,7 +264,7 @@ public class CamelYamlDSLParser extends ParserFileHelper {
 							 StringBuilder lines, int pos, boolean cleanNewLine) {
 		// If it is a multi-line, cleanNewLine decides if we convert it to one very long single line
 		for (Integer lineNo = position.getLine(); lineNo >=0; lineNo--) {
-			String tempLine = parserFileHelperUtil.getLine(textDocumentItem, lineNo);
+			String tempLine = parserFileHelperUtil.getLines(textDocumentItem, lineNo, lineNo);
 
 			for (String whenToStop : whenToStopArray) {
 				if (tempLine.stripLeading().startsWith(whenToStop)) {
--- a/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
+++ b/src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
@@ -34,9 +34,10 @@ import com.github.cameltooling.lsp.internal.instancemodel.YamlDSLModelHelper;
 public class CamelYamlDSLParser extends ParserFileHelper {
 	
 	public static final String URI_KEY = "uri";
-	public static final String REST_KEY = "rest";
 	public static final String FROM_KEY = "from";
 	public static final String TO_KEY = "to";
+
+	public static final String[] URI_SEARCH = new String[]{URI_KEY, "- " + TO_KEY};
 
 	@Override
 	public String getCamelComponentUri(String line, int characterPosition) {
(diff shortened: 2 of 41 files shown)


Delawen:
Something went wrong with the rebase from main, wait, it should be only four or five files.

Delawen:
This PR contains an "aesthetic" change in which the file `ParserFileHelperUtil.java` now takes into account that you are potentially going to return more than one line. Didn't want to completely remove the `getLine` functions, but deprecated them (to slowly remove them from everywhere) and created the new `getLines` (in plural).

Also, now we pass not only the start position but also the end position.

Delawen:
The file that has more changes is `CamelYAMLParser.java` which now has to deal with multiple lines parsing and overrides for example the `getCamelComponentUri` function, because we want to extract more than one line.

This function would probably benefit from receiving the end position too (see previous comment). But for now, we just need a position to retrieve the full YAML element with the URI.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
```diff
@@ -177,20 +191,40 @@ private boolean isStartOfUri(int separatorPos, int pos, String line) {
...
-				} else if (data.containsKey(REST_KEY)) {
-					return null;
+		StringBuilder sb = new StringBuilder();
+		extractLines(textDocumentItem, new Position(lineNumber, 0), "-", sb, 0, false);
+
+		Yaml yaml = new Yaml(new SafeConstructor(new LoaderOptions()));
+		return getCorrespondingTypeFromUnknownElement(yaml.load(sb.toString()));
+	}
+
+	private static String getCorrespondingTypeFromUnknownElement(Object yamlElement) {
```
Delawen, line 201: We need to iterate here because now, as we are multi-line, we are not sure on which level we have to search for the camel uri.

It can be a simple `to: "aws:something"` or it can be a `uri` inside a `to` element.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
```diff
@@ -220,4 +254,46 @@ private String extractUriFromYamlData(String line) {
...
+				}
+			}
+
+			// Prepend the content
+			if (cleanNewLine) {
+				// If it is multiline, we have to move the position forward
+				if (lines.length() > 0) {
+					final var tempLineStripped = tempLine.stripLeading();
+					pos += tempLineStripped.length();
+					lines.insert(0, tempLineStripped);
```
Delawen, line 286: We strip the line from all the previous whitespace because we are unifying all the string into one big string. And the trailing spaces won't make sense on this one big string.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
```diff
@@ -220,4 +254,46 @@ private String extractUriFromYamlData(String line) {
...
 	}
+
+	private int extractLines(TextDocumentItem textDocumentItem, Position position, String whenToStop,
+							 StringBuilder lines, int pos, boolean cleanNewLine) {
+		return extractLines(textDocumentItem, position, new String[]{whenToStop}, lines, pos, cleanNewLine);
+	}
+
+	private int extractLines(TextDocumentItem textDocumentItem, Position position, String[] whenToStopArray,
+							 StringBuilder lines, int pos, boolean cleanNewLine) {
+		// If it is a multi-line, cleanNewLine decides if we convert it to one very long single line
```
Delawen, line 265: Sometimes we want the YAML structure in multi-line. Sometimes we want one big string with no new lines in between. That's what cleanNewLine does.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
```diff
@@ -220,4 +254,46 @@ private String extractUriFromYamlData(String line) {
...
+		for (Integer lineNo = position.getLine(); lineNo >=0; lineNo--) {
+			String tempLine = parserFileHelperUtil.getLine(textDocumentItem, lineNo);
```
apupier, line 267: This new line is using the method getLine that is deprecated in this PR. Should it use another method? Or we shoudl not deprecate it?

apupier (changes requested):
- The completion seems to work only when we are at last line of the file
https://github.com/camel-tooling/camel-language-server/assets/1105127/08b831af-c784-4ac6-9ce5-7c090c53aad8

Delawen:
> The completion seems to work only when we are at last line of the file

Must be something else but I am investigating it.
[Screencast from 2023-11-29 10-19-51.webm](https://github.com/camel-tooling/camel-language-server/assets/726590/5c3695f5-8d51-45aa-ae83-ebeead3fc751)

src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperUtil.java
```diff
@@ -21,20 +21,40 @@
...
+		return getLines(textDocumentItem.getText(), startLine, endLine);
+	}
+
+	public String getLines(String text, int startLine, int endLine) {
+		String[] lines = text.split("\\r?\\n", endLine + 2);
+		StringBuilder sb = new StringBuilder();
+		for (int i = startLine; i <= endLine && i < lines.length; i++) {
+			sb.append(lines[i]);
+		}
+		return sb.toString();
```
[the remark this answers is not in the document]
Delawen: Makes sense. I haven't found when that would fail, but I am going to return null anyway if nothing found, just in case it is something not covered by tests (unit or my manual tests).

src/main/java/com/github/cameltooling/lsp/internal/hover/CamelKModelineHoverProcessor.java
```diff
@@ -36,7 +36,8 @@ public CamelKModelineHoverProcessor(TextDocumentItem textDocumentItem) {
 
 	public CompletableFuture<Hover> getHover(Position position, CompletableFuture<CamelCatalog> camelCatalog) {
 		int lineNumber = position.getLine();
-		CamelKModeline camelKModeline = new CamelKModeline(new ParserFileHelperUtil().getLine(textDocumentItem, lineNumber), textDocumentItem, lineNumber);
+		CamelKModeline camelKModeline = new CamelKModeline(new ParserFileHelperUtil().getLines(textDocumentItem,
+				lineNumber, lineNumber + 1), textDocumentItem, lineNumber, lineNumber);
```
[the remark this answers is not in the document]
Delawen: Now I am not sure why I did that... but anyway it works with only lineNumber in both places which would make sense. So changing it.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
Delawen: Tested with getLines and it worked. So changing it.

src/main/java/com/github/cameltooling/lsp/internal/parser/CamelYamlDSLParser.java
```diff
@@ -220,4 +254,46 @@ private String extractUriFromYamlData(String line) {
...
+
+			for (String whenToStop : whenToStopArray) {
+				if (tempLine.stripLeading().startsWith(whenToStop)) {
+					// Remove the potential '>' character of multiline
+					if (tempLine.stripLeading().substring(whenToStop.length() + 1).stripLeading().startsWith(">")) {
+						tempLine = tempLine.substring(0, tempLine.indexOf(">"));
+					}
```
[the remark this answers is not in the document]
Delawen: Adding the other multiline indicators but... the problem I see is that by the time we reach them (going backwards) we already built the StringBuilder with all the content.  Either we do two loops (one to read that indicator to know if we do or don't do the newline thing) or we just assume we are ignoring newlines because this is cameluri and that means we don't care about newlines.

src/main/java/com/github/cameltooling/lsp/internal/parser/ParserFileHelperUtil.java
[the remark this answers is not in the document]
Delawen: On a quick search I see places where returning a null may be problematic:

Like
https://github.com/camel-tooling/camel-language-server/blob/main/src/main/java/com/github/cameltooling/lsp/internal/modelinemodel/CamelKModelineOption.java#L83C1-L83C59

Are we sure we want to return null?

apupier:
status reported in specific JIRA https://issues.redhat.com/browse/FUSETOOLS2-2268

PR #1208, opened 2024-12-05
FUSETOOLS2-2508 - Configure log to use a specific Camel LSP folder

apupier:
otherwise when trying to walk files to delete old logs, depending on the content of /tmp and the rights, an exception can be thrown

It will also have the side effect to be more performant by not scanning the full /tmp folder everyday or every time the log reach 10MB

### DIFF
--- a/src/main/resources/log4j2.properties
+++ b/src/main/resources/log4j2.properties
@@ -2,7 +2,7 @@ status = warn
 name= RollingFileLogConfigCamelLSP
 
 # Log files location
-property.basePath = ${sys:java.io.tmpdir}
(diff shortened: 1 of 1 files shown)


apupier:
I locally tested that it was the root cause of the server being shutdowned by rebuilding locally by chanign the log4j configuration and using appender.rolling.policies.size.size = 10k to force a rooling update to occur. In this case, it is searchign in the basePath (previosuly /tmp) if there are logs to delete based on other confditions provided (here last modified 30 days). When walking in the /tmp, it tries to go in subfolders, and for some there was an AccessDeniedException, which was thrown to the client back and then the client was sending a shutdown request

djelinek (approved):

